PR #84, opened 2026-04-20
improve(token-alert): autoresearch evolution

aaronjmars:
### Winner: Variation D — Statistical baseline (**50.5/~52.5 weighted**)

**Thesis.** A hardcoded 10% threshold is wrong in both directions: it floods on volatile memecoins and misses genuine anomalies on low-vol majors or stablecoins. Replace it with a per-token 30-day rolling baseline of daily log-returns and daily volume, score today as a z-score, and **alert only when both price and volume are anomalous** vs that token's own history. Tier the output (CRITICAL / WATCH / quiet) by combined severity; add a 24h dedup window and multi-source fallback.

### Scoring table

| Variation | Clarity | Data | Output | Robust | Conv. | Improvement | Weighted |
|-----------|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
| A — Better inputs (markets endpoint + CoinPaprika/Binance fallback + 7d sparkline + BTC-relative) | 4 | 5 | 3.5 | 3.5 | 4 | 3.5 | **40.25** |
| B — Sharper output (severity tiers, volume-confirmed only, "top percentile of last 30d", BTC-beta) | 4 | 4 | 4.5 | 3.5 | 4 | 4 | **42.25** |
| C — More robust (fallback chain, dedup, OK/EMPTY/DEGRADED states, per-source observability) | 5 | 4 | 3 | 5 | 5 | 3 | **41** |
| **D — Statistical baseline (rethink, folds in A+B+C wins)** | 4.5 | 5 | 5 | 4.5 | 4.5 | 5 | **50.5** |

Weights: Improvement ×3, Output ×2, Clarity/Data/Robust ×1.5, Conventions ×1.

### What changed vs original

- **Threshold logic.** Dropped fixed 10% / 3× volume rule. Each token now gets a rolling 30d baseline (log-returns + volume). Alert iff `|price-z| ≥ 2.5` **and** `|vol-z| ≥ 1.5` (dual-signal confirmation). CRITICAL tier requires `|price-z| ≥ 3.0` and `vol-z ≥ 2.0`.
- **Data source.** Added CoinGecko `market_chart?days=30&interval=daily` for the baseline, kept `simple/price` for the snapshot. Added CoinPaprika and Binance `klines` as fallback chain when CoinGecko fails. Each per-token source is tracked in an observability footer.
- **Cold-start safety.** If <20 samples or `std_r == 0` (stablecoin), falls back to a `|24h| > 3%` + `vol > 2× mean` rule and logs `baseline=insufficient` rather than emitting noise.
- **Output.** Tiered notification (CRITICAL / WATCH), one line per alert with direction arrow, price, z-scores, and BTC-relative excess so the reader can tell idiosyncratic vs market-wide moves. Omits empty tier sections; caps at 8 lines; sources/baseline/suppressed footer for observability.
- **Dedup.** Greps last 2 days of logs for `TOKEN_ALERT_FIRED {sym} {tier}`; suppresses same-or-lower-tier repeats within 24h; escalates WATCH → CRITICAL if severity jumped.
- **Status codes.** Logs one of `TOKEN_ALERT_OK | TOKEN_ALERT_FIRED | TOKEN_ALERT_DEGRADED | TOKEN_ALERT_EMPTY` plus structured price/z-score line every run — future invocations can use this logged history as a secondary baseline source.

### Runners-up

- **B — Sharper output (42.25).** "Top percentile of last 30d" + severity tiers + BTC-beta adjustment. Folded into D as the tier thresholds and the BTC-relative excess column.
- **C — More robust (41).** Fallback chain + per-source observability + OK/DEGRADED/EMPTY states. Folded into D as the source fallback order, footer, and status codes.
- **A — Better inputs (40.25).** Multi-source markets endpoint + BTC-relative context. Folded into D via the `market_chart` 30d history fetch and the CoinPaprika/Binance fallbacks.

### Risks / trade-offs

- Requires 30d of daily candles per token per run (1 extra request per tracked token). Well within CoinGecko free-tier 10k/month for any reasonable tracked-tokens list.
- Statistical baseline needs history; stablecoins or brand-new tokens hit the insufficient-baseline path on day one. Handled via the fallback rule rather than silent skip.
- Dual-signal requirement is conservative by design — will miss headline-driven spikes that occur without volume confirmation. This is intentional: the previous skill's false-positive rate is the stated problem.

---
Ported from aaronjmars/aeon-tmp#41.

### DIFF
--- a/skills/token-alert/SKILL.md
+++ b/skills/token-alert/SKILL.md
@@ -1,59 +1,145 @@
 ---
 name: Token Alert
-description: Notify on price or volume anomalies for tracked tokens
+description: Notify on per-token statistically anomalous price + volume moves (not arbitrary % thresholds)
 var: ""
 tags: [crypto]
 ---
 > **${var}** — Token symbol or CoinGecko ID. If empty, checks all tracked tokens.
 
-If `${var}` is set, only check that token.
+<!-- autoresearch: variation D — statistical baseline: replace hardcoded % thresholds with per-token z-scores (price + volume), require dual-signal confirmation, tiered output -->
 
+If `${var}` is set, only check that token.
 
 ## Config
 
-This skill reads tracked tokens from a "Tracked Tokens" section in `memory/MEMORY.md`. If the section doesn't exist yet, add it to MEMORY.md or skip this skill.
+This skill reads tracked tokens from a **"Tracked Tokens"** section in `memory/MEMORY.md`. If the section is missing, add it or exit with `TOKEN_ALERT_EMPTY — no tracked tokens`.
 
 ```markdown
 ## Tracked Tokens
-| Token | CoinGecko ID | Alert Threshold |
-|-------|-------------|-----------------|
-| ETH   | ethereum    | 10%             |
-| SOL   | solana      | 10%             |
+| Token | CoinGecko ID | Notes |
+|-------|--------------|-------|
+| ETH   | ethereum     |       |
+| SOL   | solana       |       |
 ```
+
+The `Alert Threshold` column is no longer used — thresholds are now derived per-token from 30-day statistics.
 
 ---
 
-Read memory/MEMORY.md for tracked tokens and alert thresholds.
-Read the last 2 days of memory/logs/ for previous prices to detect changes.
+Read `memory/MEMORY.md` for tracked tokens. Read the last 2 days of `memory/logs/` for prior alerts (dedup).
 
-Steps:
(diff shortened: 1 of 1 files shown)


aaronjmars:
Closed per triage: replaces fixed %-threshold with z-score dual-signal (log-returns + stddev); adds CoinPaprika + Binance fallbacks unverified in Aeon. Changes *what triggers* an alert — breaks downstream consumers of the >10% rule.

PR #195, opened 2026-05-20
fix(admanage): validate TODAY_SPEND is numeric before awk spend-cap check

antfleet-ops:
Fixes H8 from the AntFleet bench run — receipt: https://github.com/AntFleet/aeon-bench/pull/4#issuecomment-4475360859

## What was wrong

The daily-spend circuit breaker fetches spend from the AdManage API and passes it to `awk` for the cap comparison:

```bash
TODAY_SPEND=$(echo "$SPEND_RESP" | jq -r '.metadata.totalSpend // 0')
if awk "BEGIN{exit !($TODAY_SPEND >= $STRICTEST_CAP)}"; then
```

`awk` coerces non-numeric values (including empty string) to `0`. If the API returns malformed JSON, or if `jq` fails and produces no output, `TODAY_SPEND` is empty. `awk` then evaluates `"" >= CAP` as `0 >= CAP`, which is false for any positive cap — the circuit breaker never trips, and ads launch against a cap that could not be verified.

The `|| echo '{}'` guard on the curl call doesn't help here: `jq -r '.metadata.totalSpend // 0'` on `{}` returns `0` (correct), but if the API returns a non-JSON response body (HTML error page, truncated response, etc.) `jq` exits non-zero and writes nothing to stdout, leaving `TODAY_SPEND` empty.

## Fix

Validate `TODAY_SPEND` against a decimal-number regex before the `awk` call. On failure, fail closed: emit a notify and exit 0, leaving pending files intact for manual inspection — the same behavior as when the cap is actually tripped.

### DIFF
--- a/scripts/postprocess-admanage.sh
+++ b/scripts/postprocess-admanage.sh
@@ -59,6 +59,15 @@ if [ "$STRICTEST_CAP" != "null" ] && [ -n "$STRICTEST_CAP" ]; then
     "$API_BASE/v1/spend/daily?startDate=$TODAY&endDate=$TODAY" \
     -H "$auth_hdr" || echo '{}')
   TODAY_SPEND=$(echo "$SPEND_RESP" | jq -r '.metadata.totalSpend // 0')
+  # Guard: if the API response was malformed or empty, jq may produce a
+  # non-numeric value. awk treats "" as 0, which would silently bypass the cap.
+  # Fail closed instead — block the launch until spend can be verified.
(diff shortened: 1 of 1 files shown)


imancipate:
**Triage:** OUT-OF-SCOPE — touches `scripts/postprocess-admanage.sh`, a protected post-process script that external contributors cannot modify directly.
For changes here, please open an issue describing the goal and a maintainer will land the change. Leaving open and labelled `triage:out-of-scope` for maintainer attention; the issue is the right venue for the actual fix.

PR #200, opened 2026-05-21
Add opt-in Fleet Watcher authorization layer (preflight + postflight)

opengrid1:
## What

Adds two new steps to `.github/workflows/aeon.yml` that wrap the existing `Run` (Claude Code execution) step with synchronous authorization against a self-hosted [Fleet Watcher](https://github.com/yourorg/fleet-watcher) control plane.

| Step | Position | Effect |
|---|---|---|
| **Fleet Watcher preflight** | Before `Run` | `POST /api/aeon/preflight`. If Fleet returns `allow: false`, the workflow exits 1 — Claude never starts, no side-effects. |
| **Fleet Watcher postflight** | After `Run`, `if: always()` | `POST /api/aeon/postflight`. Records outcome (`ok` / `error` / `blocked`) and feeds Fleet's source→sink taint chain detection. |

## Backward compatible

Both steps gate on `env.FLEET_ENDPOINT != '' && env.FLEET_TOKEN != ''`. Without those repo secrets, both steps no-op. Every existing AEON fork keeps running exactly as before.

## Safety properties

- **Fail-closed**: when secrets are set but Fleet is unreachable, preflight exits non-zero. Claude never runs.
- **Postflight always runs** (`if: always()`): failed and blocked skills are still recorded — failure itself is signal for taint analysis.
- **Idempotent within an attempt**: each step sends `clientOpId=${RUN_ID}-${ATTEMPT}-{pre,post}` so step re-evaluation dedupes to a single verdict / event server-side.
- **No new permissions or runners**: only `curl` + `jq` (already available on `ubuntu-latest`).
- **Secrets at workflow-env scope**: hoisted to the top-level `env:` block so the `if:` clauses can actually see them (step-level env is not visible in `if:`).

## How to enable on a repo

1. Stand up Fleet Watcher.
2. Mint an agent token: `POST /api/aeon/register` with `{"name":"my-aeon-fork","environment":"production"}`.
3. Add two repo secrets:
   - `FLEET_ENDPOINT` = base URL of your Fleet
   - `FLEET_TOKEN` = the `agnt_…` returned by register
4. Define policies in the Fleet dashboard.

Full setup steps and rationale are in the new "Fleet Watcher" section added to `README.md`.

### DIFF
--- a/.github/workflows/aeon.yml
+++ b/.github/workflows/aeon.yml
@@ -61,6 +61,11 @@ permissions:
 env:
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
   NODE_OPTIONS: '--no-deprecation'
+  # Fleet Watcher opt-in secrets — hoisted to workflow-level so that
+  # `if: env.FLEET_ENDPOINT != ''` in the preflight/postflight steps can
+  # actually see them. (Step-level env is NOT available in step `if:` clauses.)
+  FLEET_ENDPOINT: ${{ secrets.FLEET_ENDPOINT }}
+  FLEET_TOKEN:    ${{ secrets.FLEET_TOKEN }}
 
 concurrency:
   group: aeon-${{ inputs.skill || github.event.action || github.run_id }}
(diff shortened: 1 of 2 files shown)


imancipate:
For changes here, please open an issue describing the goal (the design and safety properties of the Fleet Watcher pre/postflight wrap are well laid out and would convert cleanly into one) and a maintainer will land the change. Closing-as-not-planned to keep the queue tidy; the issue is the right venue.

imancipate:

Issue #184, opened 2026-05-18
AntFleet two-model review surfaced 12 findings on recent merged PRs


antfleet-ops:
Hi @aaronjmars,

We benchmark agent-framework repos on [AntFleet](https://antfleet.dev) using a two-model unanimous review gate (Claude Opus 4.7 + GPT-5). On 2026-05-18 we ran the gate against 15 cherry-picked PRs from this repo via a public mirror ([antfleet/aeon-bench](https://github.com/antfleet/aeon-bench)). The gate posted 27 findings as SHA-pinned receipts. 14 of the 15 reviewed PRs are now merged into your main, so most of the flagged code is live.

This issue is informational — it is an index of the findings, each linked to its public receipt. You are under no obligation to act on any of them. Happy to send focused PRs on any subset if useful — just reply.

### C1 · Unauthenticated endpoint can trigger GitHub Actions via gh CLI
- **File**: `dashboard/app/api/skills/[name]/run/route.ts:11-18`
- **Source PR**: #158 (merged 2026-05-06)
- **Receipt**: https://github.com/AntFleet/aeon-bench/pull/1#issuecomment-4475345067

### C2 · Missing authentication/authorization on secrets API
- **File**: `dashboard/app/api/secrets/route.ts:59-119`
- **Source PR**: #150 (merged 2026-05-03)
- **Receipt**: https://github.com/AntFleet/aeon-bench/pull/3#issuecomment-4475352768

### C3 · Missing authentication/authorization on secret-management endpoints
- **File**: `dashboard/app/api/secrets/route.ts:95-145`
- **Source PR**: #169 (merged 2026-05-14)
- **Receipt**: https://github.com/AntFleet/aeon-bench/pull/25#issuecomment-4475682712

### H1 · v4-readiness manifest references files not included in the skill's read set
- **File**: `skills/v4-readiness/SKILL.md`
- **Source PR**: #160 (merged 2026-05-07)
- **Receipt**: https://github.com/AntFleet/aeon-bench/pull/12#issuecomment-4475654244

### H2 · Slack bot-message filter is inverted: `BOT_ID = "null"` string check rejects real users
- **File**: `.github/workflows/messages.yml:105-117`
- **Source PR**: #29 (merged 2026-04-13)
- **Receipt**: https://github.com/AntFleet/aeon-bench/pull/2#issuecomment-4475356633

### H3 · Undefined `FORK_DEFAULT_BRANCH` used to fetch aeon.yml from the fork
[...]

PR #203, opened 2026-05-21
fix(fleet-state): create .bak before write + validate TMP before promote (closes AntFleet H4)

antfleet-ops:
## Summary

The persist-state step in `skills/fleet-state/SKILL.md` declares *"keep one `.bak` rolling. If `jq empty` fails after write → restore from `.bak`"* — but nothing in the step ever creates the `.bak`. The sequence is `mv $TMP ...json` first, then `jq empty` validation on the live file, with rollback attempting `cp ...json.bak ...json`. On a failed validate, the rollback `cp` silently fails (no source) and the operator is left with a corrupt `fleet-state.json` and no recovery path — and since the live file was already overwritten, the previous week's state is unrecoverable.

## Evidence

This issue was flagged by AntFleet's two-model unanimous review gate on a benchmark replay of #168:

- AntFleet receipt: https://github.com/AntFleet/aeon-bench/pull/24#issuecomment-4475370933
- Original PR (merged 2026-05-14): https://github.com/aaronjmars/aeon/pull/168
- Specific finding: **High · Data-loss** — State file rollback uses non-existent `.bak`, can cause data loss on write failure
- Location: `skills/fleet-state/SKILL.md:291-306` (step 11)

Indexed from Issue #184 (closed) as **H4**.

The finding was an agreement between two independent frontier reviewers (Claude Opus 4.7 + GPT-5). See https://antfleet.dev for context on the review methodology.

## Proposed fix

Two adjustments to the persist sequence, both minimal:

1. Add a `cp …json …json.bak` step **before** the write attempt, so the rollback path always has a non-empty backup. Runs every persist iteration as a rolling backup.
2. Move the `jq empty` check from *after* the `mv` to *before* the `mv` — the candidate JSON is validated in the temp file, and only promoted to the live file if valid. If it's invalid, the temp file is discarded and the live file is restored from `.bak` defensively. The error log is preserved.

Net effect: a malformed jq output now leaves the live file at the prior good state instead of replacing it with a corrupt blob.

## Notes for the maintainer

- This is a responsible-disclosure PR; you are under no obligation to merge.
- If you prefer a different approach (or have already addressed this in a separate WIP), feel free to close — the receipt stays on the bench PR regardless.
- Happy to revise the patch to match your project conventions if requested.

### DIFF
--- a/skills/fleet-state/SKILL.md
+++ b/skills/fleet-state/SKILL.md
@@ -288,6 +288,12 @@ Cap the article at ~400 lines. If any section's bullet list exceeds the cap, tri
 ### 11. Persist state
 
 ```bash
+# Roll the .bak forward BEFORE we touch the live file. This is the only
+# place that creates the backup; without this line the rollback path
+# below (cp .bak ...) would have nothing to restore from on a corrupt
+# write — both the live file and the backup would be lost.
+[ -f memory/topics/fleet-state.json ] && cp memory/topics/fleet-state.json memory/topics/fleet-state.json.bak
+
 TMP=$(mktemp)
 jq --arg ts "$(date -u +%FT%TZ)" \
    --arg today "$(date -u +%F)" \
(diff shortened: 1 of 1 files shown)


imancipate:
**Triage:** ACCEPTED — clean against the contribution rubric (scope ✓ / format ✓ / originality n/a / size ✓).

PR #223, opened 2026-05-24
fix(run): replace fragile tr-d quoting in GATEWAY provider parsing

antfleet-ops:
## Summary

The GATEWAY provider extraction line used a malformed quoting sequence:

```bash
GATEWAY=$(... | tr -d ' "'"'"' || echo "direct")
```

The `tr -d` argument mixes single and double quotes in a way where the trailing `"` bleeds into the `||` fallback, making shell parsing fragile and causing the Run step to fail when `gateway.provider` is set in `aeon.yml`.

**Fix:** Replace with two clean `sed` calls using double-quote context:
```bash
GATEWAY=$(... | sed 's/.*provider:[[:space:]]*//' | sed "s/[\"' ]//g" || true)
```
The `|| true` is safe because the fallback is already handled by `GATEWAY="${GATEWAY:-direct}"` on the next line.

**Severity:** High — a syntax error here aborts the shell step and prevents model routing from being determined.

Found by AntFleet two-model consensus review (claude-opus-4-7 + gpt-5, unanimous).
Benchmark: https://github.com/AntFleet/aeon-bench/pull/31


imancipate:

imancipate:

PR #224, opened 2026-05-24
fix(notify): write dedup hash only after successful delivery

antfleet-ops:
## Summary

The dedup hash was written to `.notify-sent-hashes` **before** attempting delivery:

```bash
printf '%s\n' "$HASH" >> "$HASH_FILE"   # written here

# Always save to .pending-notify/ ...
DELIVERED=false
# ... attempt delivery ...
if [ "$DELIVERED" = "true" ]; then
  rm -f ".pending-notify/${TS}.md"
fi
```

If all transports fail inside the sandbox, the message is saved to `.pending-notify/` AND the hash is already in the dedup file. The post-run *Send pending notifications* step then skips it because the hash matches — **silent permanent message loss**.

**Fix:** Move the `printf` inside the `DELIVERED=true` block so only actually delivered messages are recorded as sent.

**Severity:** Medium — data-loss path triggered when sandbox blocks curl (the documented reason the `.pending-notify/` fallback exists).

Found by AntFleet two-model consensus review (claude-opus-4-7 + gpt-5, unanimous).
Benchmark: https://github.com/AntFleet/aeon-bench/pull/31


imancipate:

imancipate:

aaronjmars:
Thanks for digging into this — but I'm going to close it, because moving the hash write to after `DELIVERED=true` actually reintroduces the duplicate send it's meant to prevent.

`.notify-sent-hashes` is truncated at the start of each run, and the per-run hash dedup exists to stop the **same message being re-queued** if `./notify` is called again. The reliable delivery path isn't the immediate curl (which is *expected* to fail under the sandbox) — it's the `.pending-notify/<ts>.md` file the post-run step picks up.

With the hash written only on immediate-delivery success, a sandboxed run never records the hash, so a second identical `./notify` call writes a **second** pending file → the post-run step then delivers the same message twice. The retry-on-failure you're after is already covered by the pending-notify fallback, so gating the hash on delivery isn't needed and regresses dedup in the common case.

Happy to look at a version that keeps the dedup write up-front if you hit a concrete case it doesn't cover.

PR #225, opened 2026-05-24
docs(ecosystem): list LiquidPad

liquidpadbot:
Adds a row to ECOSYSTEM.md for LiquidPad — an independent token launchpad on Base built on Liquid Protocol's open primitives.

## Why this row

ECOSYSTEM.md (introduced in #220) curates **products built with Aeon**, separate from:
- Community skill packs (`README.md`) — LiquidPad already listed there as `aeon-skill-pack-liquidpad` via #218
- SHOWCASE.md — for forks running stock skills

LiquidPad fits ECOSYSTEM.md's guidelines because it's a product:
- Uses Aeon's skill format
- Ships an installable pack co-authored on `0eed836`
- Publicly identifies as building on the Aeon ecosystem

## Row inserted

Alphabetically between **LawbWorld** and **Liq** (per the file's alphabetization rule):

```
| LiquidPad | [@LiquidPadBot](https://x.com/LiquidPadBot) · [liquidpad.site](https://www.liquidpad.site) |
```

## Compliance with file guidelines

- ✅ Alphabetized
- ✅ One row per project
- ✅ X handle + website link
- ✅ Project publicly identifies as built on Aeon (see [/built-on-liquid](https://www.liquidpad.site/built-on-liquid) for full attribution)
- ✅ Already shipping a community skill pack (cross-listed in README per the ECOSYSTEM.md guideline)

Happy to revise wording, link order, or any rendering detail if it doesn't fit.

PR #231, opened 2026-05-25
feat(skills): add liquidpad-launch

liquidpadbot:
Adds a `liquidpad-launch` skill — drop-in primitive for any aeon agent to ship a token through LiquidPad on Base via the public API.

## What it does

Deploys a token using LiquidPad's `/agent/run-once` endpoint. The contract-enforced fee split is the part most agents reinvent badly:

- 80% creator (no claim limits, no time locks)
- 15% buys $LPAD on market and burns it
- 5% buys $LIQ (Liquid Protocol)

Every shipped token is ERC-8004 stamped (agent #50962, registered onchain on Base).

## Safety policy (mirrors the auto-merge skill style)

A launch fires only when every gate passes:

- API key gate (`LIQUIDPAD_API_KEY`)
- Concept fields valid (name 1–32, symbol 2–10 alnum, theme ≤ 200)
- No same-day duplicate symbol (24h window via `memory/topics/liquidpad-state.json`)
- No banlist match (`memory/topics/liquidpad-banlist.md`)
- Owner wallet resolved from `OWNER_WALLET` env or `memory/watched-repos.md`
- Daily cap (`MAX_LAUNCHES_PER_DAY`, default 1)
- Dry-run mode (`LIQUIDPAD_DRY_RUN=1`)

## Failure modes (structured)

`DEPLOY_AUTH_FAIL` / `DEPLOY_RATE_LIMIT` / `DEPLOY_UPSTREAM_ERROR` / `VERIFY_FAIL` / `DEPLOY_PARTIAL` — each surfaces as a distinct verdict so logs stay debuggable.

## Constraints

- Never deploys with placeholder `ownerAddress`
- Never re-deploys the same symbol within 24h (Uniswap V4 confusion)
- Never calls `/agent/run-once` more than once per invocation
- API key never persisted to memory or logs

## Companion to ECOSYSTEM.md entry

LiquidPad has been listed in ECOSYSTEM.md since #225 — this PR ships the actual primitive so the listing has substance behind it.

## Files

- `skills/liquidpad-launch/SKILL.md` (new)
- `skills.json` (regenerated via `./generate-skills-json --pretty` — 156 → 157 skills)

## References

- LiquidPad: https://www.liquidpad.site
- Ecosystem attribution + live contributions: https://www.liquidpad.site/ecosystem
- Public API: https://api.liquidpad.site
- Skill manifest (LLM-readable): https://www.liquidpad.site/agent-owner-launch-skill.md
- ERC-8004 record: https://8004agents.ai/base/agent/50962


aaronjmars:
Thanks for the `liquidpad-launch` skill — the skill itself looks good: egress is limited to your own `api.liquidpad.site`, the API key only goes in the `x-api-key` header and isn't persisted, and the safety gating (API-key gate, owner-wallet validation, banlist, 24h dedupe, daily cap, dry-run) is solid.

One blocker before I can merge: the `skills.json` in this PR appears to have been regenerated from a shallow/squashed checkout, so it churns ~156 unrelated entries — most entries' `sha` collapse to a single commit (`d72ace6`) and `updated`/`schedule`/`category` get rewritten on many rows (the new entry's own `sha`/`updated` are also empty). Install targets are untouched so it's not dangerous, just noisy/inaccurate metadata across rows unrelated to this PR.

Could you either regenerate `skills.json` from a full-history checkout (so per-skill `sha`/`updated` resolve correctly), or scope the diff to only the new `liquidpad-launch` entry? Happy to merge once `skills.json` is limited to the single addition.

Minor: the curl body around line 61 interpolates `$VAR` unquoted into JSON (`-d "{\"vibe\":\"$VAR\"}"`) — worth quoting/escaping for robustness.

imancipate:
**Triage:** DEFER — depends on an external secret (`LIQUIDPAD_API_KEY`) that the maintainer hasn't provisioned, and the deploy path calls the API directly via `curl` with no `scripts/prefetch-*.sh` (so it won't run under the GitHub Actions sandbox as written).
This PR is sound but needs maintainer attention before it can move (secret provisioning + a prefetch/postprocess shim for the auth'd call). Leaving open for @aaronjmars to pick up on the next review pass.

aaronjmars:
Thanks @liquidpadbot — the `skills.json` scoping and the `jq`-escaped body both look good now, and as I said the safety gating (api-key gate, owner-wallet validation, banlist, 24h dedupe, daily cap, dry-run) is solid. I'm open to landing this; one real blocker remains, and it's the sandbox issue @imancipate flagged.

**The problem.** aeon skills run in a sandbox that blocks outbound `curl` with a secret in the header — the auth header can't expand `$LIQUIDPAD_API_KEY` at skill-execution time. (Same reason `refresh-x` never curls `api.x.ai` from the skill body; see the note at the bottom of `skills/refresh-x/SKILL.md`.) So the inline `curl ... -H "x-api-key: $LIQUIDPAD_API_KEY"` calls in steps 1, 5, and 6 won't run as written.

**The convention.** Authed network I/O is split into two shims that run *outside* the sandbox with the secret injected:

- **`scripts/prefetch-*.sh`** — runs *before* the agent, for authed **reads**; writes results to a cache file the skill then reads. Model: `scripts/prefetch-xai.sh` → `.xai-cache/<name>.json`, consumed via "Path A — prefetched cache".
- **`scripts/postprocess-*.sh`** — runs *after* the agent, for authed **writes** / side-effecting calls. Model: `scripts/postprocess-devto.sh` reads `.pending-devto/*.json` and POSTs each with `DEVTO_API_KEY`.

**How it maps for liquidpad-launch:**

1. **Concept** (`POST /agent/concept`, authed read) → add a `liquidpad-launch)` case to **`scripts/prefetch-liquidpad.sh`** that fetches the concept and writes `.xai-cache/liquidpad-concept.json`. The skill reads that file instead of curling in step 1.
2. **Deploy** (`POST /agent/run-once`, authed + side-effecting) → this is the key one. The skill body runs *all* the safety gates, and **only if every gate passes** writes the deploy request body to `.pending-liquidpad/<symbol>.json` plus its `go`/`skip` verdict to the log. Then **`scripts/postprocess-liquidpad.sh`** — running outside the sandbox with `LIQUIDPAD_API_KEY` — reads `.pending-liquidpad/*.json`, performs the single POST, parses `result.token.address` / `result.txHash`, and removes the payload on success (exactly like `postprocess-devto.sh`). Honor `LIQUIDPAD_DRY_RUN` inside the script too (write the body, skip the POST).
3. The skill body must **never** curl `api.liquidpad.site` with the key directly — drop the inline curls and add an `## Environment / sandbox` note documenting the cache + `.pending-liquidpad/` handoff, the way `refresh-x` does.

Two things to keep correct in the split:
- **Daily cap / dedupe** are evaluated in the skill (it decides whether to emit a payload at all), so a payload only lands in `.pending-liquidpad/` when under cap and non-duplicate — don't rely on the script to re-check.
- **`OWNER_WALLET`** resolution stays in the skill body (it's not secret); the resolved address goes into the payload so postprocess just passes it through.

On my side: I'll provision **`LIQUIDPAD_API_KEY`** as a workflow secret once the prefetch/postprocess shims are in, so the authed steps have the key where they actually run. Push those and I'll re-review and merge. Appreciate the iteration.

imancipate:
**Triage:** OUT-OF-SCOPE — touches `scripts/prefetch-liquidpad.sh` and `scripts/postprocess-liquidpad.sh`, which the contribution rubric treats as protected paths that external contributors cannot land directly.

Context: yesterday's `**Triage:** DEFER` flagged the missing `scripts/prefetch-*.sh` shim — that was a genuine blocker, and adding it was the right response. The rubric tension is that the prefetch/postprocess shims themselves need a maintainer to land (same reason: they execute outside the Claude sandbox with `LIQUIDPAD_API_KEY` in scope). The skill body (`skills/liquidpad-launch/SKILL.md`) and `skills.json` regeneration are fine; the two `scripts/` files are the bit that needs maintainer hands.

One path forward without a split-PR: a maintainer (cc @aaronjmars) lands the two `scripts/` files directly on `main`, then this PR rebases to drop them and ships just the skill + registry — at which point it cleanly re-triages as ACCEPTED. Alternatively, restructure the skill to fail-soft when `LIQUIDPAD_API_KEY` is unset (no sandbox bypass needed), so the skill can land standalone and the deploy path activates later.

Leaving the PR open and labelled `triage:out-of-scope` — not closing, because the contribution is well-scoped and the blocker is on the maintainer side, not the author's.

aaronjmars:
Landed the prefetch + postprocess shims directly on `main` in #260 — verbatim copies of the two files from this PR's head (`d67947f`), so no review surface change.

To unblock this PR:

1. Rebase your branch onto current `main` (which now contains `scripts/prefetch-liquidpad.sh` and `scripts/postprocess-liquidpad.sh`).
2. Drop those two files from this PR's diff — the rebase should turn them into no-ops since the file contents match.
3. Push the rebased branch.

After that, the PR will be down to just `skills/liquidpad-launch/SKILL.md` + the scoped `skills.json` entry, which @imancipate's rubric clears as `ACCEPTED`.

I'll provision `LIQUIDPAD_API_KEY` as a workflow secret in parallel so the authed paths have the key where they actually run. Re-tagging for re-review once you've pushed the rebase.

cc @imancipate

imancipate:
Thanks @liquidpadbot. The previous OUT-OF-SCOPE block is resolved — `scripts/prefetch-liquidpad.sh` and `scripts/postprocess-liquidpad.sh` landed via #260 yesterday, so this PR now only touches `skills/liquidpad-launch/SKILL.md` + `skills.json`. A maintainer review pass will follow.

aaronjmars (changes requested):
Solid design overall — the prefetch/postprocess shim split keeps `LIQUIDPAD_API_KEY` out of the sandbox, and the conservative defaults (`runImmediately: false`, daily cap, banlist) are the right call for a skill that can deploy onchain. Two things before merge:

**1. Owner-wallet resolution can route creator fees to the wrong address.** The safety policy says the wallet comes from `OWNER_WALLET` or a `## Owner Wallet` section, but step 3 greps the whole file for the first match:

```bash
OWNER=${OWNER_WALLET:-$(grep -oE '0x[a-fA-F0-9]{40}' memory/watched-repos.md | head -1)}
```

If `watched-repos.md` has any other `0x…` address earlier in the file (a watched contract/wallet), creator fees route there — real money, and it contradicts the stated policy. Please pin to the labeled `## Owner Wallet` section, or require `OWNER_WALLET` outright.

**2. `skills.json` is ~78 lines of unrelated churn.** Most of the diff swaps the ASCII-escaped em-dashes (the `\uXXXX` JSON form) for literal `—` characters across existing skill descriptions — a serializer artifact from regenerating without ASCII output. The real change is the single `liquidpad-launch` entry. Please regenerate with the repo's canonical encoding (jq's default ASCII `\uXXXX`) so the diff is just the new entry plus `total`/`generated`. As-is it conflicts with any other `skills.json` change and the encoding flips back on the next regen. (Minor: the entry has `"files": 0` despite shipping a `SKILL.md`.)

Happy to merge once the wallet resolution is pinned and the diff is back down to the single entry.

aaronjmars (changes requested):
Both items from the last pass are correctly addressed — thanks for the quick turnaround.

- **Owner-wallet resolution** — pinned to the labeled `## Owner Wallet` section via the `awk` range, env still wins, and the regex check now rejects malformed values with `SKIP:invalid-owner-wallet`. The "never pick the first `0x` elsewhere in the file" rule is stated in the safety-policy bullet too. 👍
- **`skills.json`** — back to just the single `liquidpad-launch` entry plus the `generated`/`total` header lines, `"files": 1`, and the `sha` matches the SKILL.md blob hash. The em-dash churn is gone.

One new thing before merge — the wallet fix introduced an **unbalanced code fence**. SKILL.md now has an odd number of ` ``` ` lines: the step 3 `bash` block opens at line 78 and is never closed. The next fence (line 95, ` ```bash `) carries an info string, so GitHub's renderer treats it as content rather than a closer — the block runs all the way to line 117, swallowing step 4's prose and the whole `jq` payload into one code blob and breaking the numbered list.

Fix is one line — close the block after the `invalid-owner-wallet` check:

```bash
   if ! [[ "$OWNER" =~ ^0x[a-fA-F0-9]{40}$ ]]; then
     echo "SKIP:invalid-owner-wallet"; exit 0
   fi
```                <-- add this closing fence

4. **Build the deploy payload** ...
```

Once the fence is closed this is good to merge. (Non-blocking nits if you're already in there: `mcEth` is passed as a string via `--arg`; the API likely wants a number.)

imancipate:
Thanks @liquidpadbot. A maintainer review pass will follow; in the meantime no changes requested from this triage layer.

PR #269, opened 2026-05-29
feat(skills): add Hound onchain investigation skills (rug-scan, contract-audit, wallet-profile, deployer-trace, tx-explain, holder-concentration)

houndflow:
Adds six **onchain investigation** skills for Base — the security/forensics gap in the current crypto skill set, which today is monitoring/market only (`on-chain-monitor`, `wallet-digest`, `defi-monitor`, `token-*`, `monitor-runners`). Those answer *"what moved?"*; these answer *"is this safe, and who is behind it?"*

## Skills

| Skill | What it does |
|-------|--------------|
| `rug-scan` | Rug-pull risk verdict for a token — ownership, mint/freeze, mutable fees, LP lock, top-holder concentration → one weighted score |
| `contract-audit` | Capability matrix — verification, proxy/upgradeability (EIP-1967), owner/admin roles, mint/freeze/pause/drain powers and whether each is live & non-renounced |
| `wallet-profile` | Behavioral profile — age, activity class (bot/whale/sniper/trader/holder/deployer), funding source, top counterparties, risk flags |
| `deployer-trace` | Maps every contract from a deployer, links reused templates, flags serial ruggers |
| `tx-explain` | Plain-English transaction decode — method, token movements, counterparties, suspicious-approval flags |
| `holder-concentration` | Distribution analysis — top-N share, HHI, LP/lock/burn exclusions, whale clustering |

## Keyless by design

Every skill runs on **public endpoints** — Etherscan v2 unified (`chainid=8453`) + Base RPC (`mainnet.base.org`). An optional `ETHERSCAN_API_KEY` only raises the rate limit and is appended to the URL, never sent in a header. **No maintainer-provisioned secret is required** to run any of these.

## Conventions followed

- Only adds `skills/<slug>/SKILL.md` (×6). **No protected paths touched** — no `scripts/prefetch-*` or `postprocess-*`; the keyless public APIs work directly with the WebFetch sandbox fallback already used by `on-chain-monitor`.
- `skills.json` patched **minimally**: just the `generated` timestamp + `total` (159→165) and 6 entries appended. No existing rows reordered or modified.
- Registered in `aeon.yml` **disabled** (`workflow_dispatch`, `var: ""`) — operator enables explicitly. These are on-demand: `var` is the target address / tx hash.
- Each skill has a `## Sandbox note` (WebFetch fallback), a `## Constraints` block (no trade advice; no invented data; verdicts come from the documented score tables), an audit-trail log to `memory/logs/`, and notifies only when a finding warrants it.
- `generate-skills-json` gets a one-line category mapping so the six slugs regenerate as `crypto`.

## Safety

Read-only — no transaction signing, no `onchain_writes`. All fetched content (source code, addresses, symbols) is treated as untrusted and never interpolated into shell beyond the single quoted target var.

Happy to split into per-skill PRs if you'd prefer to land them incrementally — starting with `rug-scan`.

### DIFF
--- a/aeon.yml
+++ b/aeon.yml
@@ -36,6 +36,14 @@ skills:
   distribute-tokens: { enabled: false, schedule: "workflow_dispatch", var: "" } # on-demand — var: distribution list label
   defi-overview: { enabled: false, schedule: "0 12 * * *" }
 
+  # --- Hound onchain investigation (on-demand; var = address / tx hash) ---
+  rug-scan: { enabled: false, schedule: "workflow_dispatch", var: "" } # var: token address — rug-pull risk verdict
+  contract-audit: { enabled: false, schedule: "workflow_dispatch", var: "" } # var: contract address — powers/ownership/proxy audit
+  wallet-profile: { enabled: false, schedule: "workflow_dispatch", var: "" } # var: wallet address — behavioral profile
+  deployer-trace: { enabled: false, schedule: "workflow_dispatch", var: "" } # var: deployer/token address — serial-rug detection
+  tx-explain: { enabled: false, schedule: "workflow_dispatch", var: "" } # var: tx hash — plain-English decode
+  holder-concentration: { enabled: false, schedule: "workflow_dispatch", var: "" } # var: token address — distribution/HHI analysis
+
   monitor-polymarket: { enabled: false, schedule: "30 12 * * *", model: "claude-sonnet-4-6" } # monitor specific markets for 24h price moves and volume changes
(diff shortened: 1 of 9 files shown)


imancipate:
**Triage:** DEFER — size (additions+deletions = 696, above the 500-line first-touch threshold without a `large-ok` label).
This PR is sound but exceeds the per-PR size bound this triage layer waves through. Two paths forward:
1. A maintainer applies the `large-ok` label to override the size gate.
2. Split into per-skill PRs as you offered — starting with `rug-scan` lets `contract-audit` / `wallet-profile` land in parallel rather than serialized behind one large review.
Leaving open and labelled `triage:deferred`; @aaronjmars will pick this up on the next maintainer review pass.

PR #282, opened 2026-05-29
feat(skills): add honeypot-check onchain investigation skill

houndflow:
Adds the `honeypot-check` skill — extends the Hound onchain-investigation set merged in #269.

## What it does
`honeypot-check` — detects **un-sellable / restricted (honeypot)** tokens on Base. A honeypot lets you buy but blocks selling; this flags that **before** you're trapped.

## How it works (keyless, zero-risk)
1. Confirms the address is a contract (`eth_getCode`).
2. Samples a real current holder from recent `Transfer` logs (adaptive block range to respect public-RPC result caps).
3. `eth_call` `transfer(recipient, amount)` with `from` = that holder — a pure dry-run (no state change, no funds).
4. Revert / `false` return → `LIKELY_HONEYPOT`; success → `SELLABLE`; no holder → `INCONCLUSIVE`.

Runs entirely on the Base RPC — **no explorer/API key** (`BASE_RPC_URL` optional).

## Conventions
- Only adds `skills/honeypot-check/SKILL.md`. **No protected paths.** WebFetch sandbox fallback documented.
- `skills.json` patched minimally (header + 1 entry); `aeon.yml` entry disabled (`workflow_dispatch`, var = token address).
- Read-only / `eth_call` only — never sends a transaction. `## Sandbox note` + `## Constraints` included (notably: this checks sell-*restriction*, not tax rate).


imancipate:
Thanks @houndflow. A maintainer review pass will follow; in the meantime no changes requested from this triage layer.

PR #287, opened 2026-05-29
feat(skills): add investigation-report composite skill

houndflow:
Adds the `investigation-report` skill — a composite that ties together the Hound onchain-investigation set merged in #269.

## What it does
The **"tell me everything about this token"** skill. Instead of running four checks by hand, it composes them into one structured report with a one-line verdict on top:

1. **Rug Scan** — red-flag score (`LOW`/`ELEVATED`/`HIGH`)
2. **Contract Audit** — verified? proxy? owner powers / renounced?
3. **Deployer Trace** — who deployed it + their history
4. **Holder Concentration** — top-holder / whale risk

## How it works
- Runs the four sub-investigations and merges their output under one heading with an **at-a-glance** summary (`Rug risk: … · Source: … · Top holder: …`).
- **Degrades gracefully:** each section is independent, so a key-gated or empty section is marked `unavailable` — explicitly *not* treated as a pass — without aborting the report.
- Emits **one** consolidated notify when the composite is concerning (rug `HIGH`, or `ELEVATED` + another flag), rather than four separate alerts.

Keyless core on the Base RPC; a `BASESCAN_KEY` deepens verified source, deployer history and the holder list.

## Conventions
- Only adds `skills/investigation-report/SKILL.md`. **No protected paths.** WebFetch sandbox fallback documented (incl. never echoing the optional key).
- `skills.json` patched minimally (header + 1 entry); `aeon.yml` entry disabled (`workflow_dispatch`, var = token address).
- Read-only — no transactions. `## Sandbox note` + `## Constraints` included (aggregator accuracy is bounded by sub-checks; unavailable ≠ safe; heuristic, not advice; no double-notify).

Note: the security scanner flags only the usual benign `` `${var}` ``/`curl "$RPC"` patterns shared by the merged #269 skills — same false-positives, reviewed manually.


imancipate:
Thanks @houndflow. A maintainer review pass will follow; in the meantime no changes requested from this triage layer.

PR #414, opened 2026-06-10
feat(gateway): add BEAMR as an LLM gateway

0xShak:
## What

Adds **[BEAMR](https://askbeamr.com)** to Aeon's LLM gateway roster (`direct | bankr | openrouter | usepod | surplus | venice | **beamr**`).

BEAMR is an OpenAI-compatible **inference router**: for each request it classifies difficulty, routes to the **cheapest capable provider**, and settles that single call in **USDC on Base over x402**. It's the same SIDECAR shape as Surplus/Venice, so it rides the existing `claude-code-router` bridge — no new payment code in Aeon.

## Why it's a little different from the other gateways

BEAMR is **self-hostable** — operators run their own router — so the routing trigger is the **deployment URL** (`BEAMR_BASE_URL`), not an API-key prefix. It behaves like Venice/UsePod in the dashboard (dropdown, no prefix). `BEAMR_API_KEY` is optional (only if your deployment is auth-gated); `BEAMR_MODEL` defaults to `auto` so BEAMR's own policy layer picks the model per call.

## Changes

- **`scripts/llm-gateway.sh`** — new `beamr)` case (SIDECAR tier via `start_ccr_sidecar`, pointing ccr at `${BEAMR_BASE_URL}/api/v1/chat/completions`).
- **Dashboard** — `GatewayProvider` enum + `GATEWAY_PROVIDERS`; auth-modal gateway detection; `BEAMR_BASE_URL` secret descriptor and `GATEWAY_SECRETS` mapping (setting it flips `aeon.yml` to `beamr`, removing it reverts to `direct`).
- **Workflows** (`aeon.yml`, `messages.yml`) — pass `BEAMR_BASE_URL` / `BEAMR_API_KEY` / `BEAMR_MODEL` into the run steps; exclude the BEAMR gateway secret from per-skill secret extraction.
- **README** — gateway table row + model-override notes.

## Compatibility

No behavior change for any existing gateway — purely additive. When `BEAMR_BASE_URL` is unset the `beamr` route is never selected, and `require_secret` fails the step cleanly if `gateway.provider: beamr` is set without it (mirrors the other gateways).

## Validation

- `bash -n scripts/llm-gateway.sh` ✅
- YAML parse of both workflows ✅
- Mirrors the existing `surplus)`/`venice)` sidecar pattern verbatim.

### DIFF
--- a/.github/workflows/aeon.yml
+++ b/.github/workflows/aeon.yml
@@ -151,7 +151,7 @@ jobs:
           SECRETS=$(grep -oE '\$\{?[A-Z][A-Z0-9_]{2,}\}?' "$SKILL_FILE" 2>/dev/null | \
             sed 's/[${}]//g' | \
             grep -E '_(API_KEY|KEY|TOKEN|SECRET|WEBHOOK_URL)$' | \
-            grep -vE '^(GITHUB_TOKEN|GH_TOKEN|ANTHROPIC_API_KEY|CLAUDE_CODE_OAUTH_TOKEN|BANKR_LLM_KEY|OPENROUTER_API_KEY|USEPOD_TOKEN|VENICE_API_KEY|SURPLUS_API_KEY|TELEGRAM_|DISCORD_|SLACK_)' | \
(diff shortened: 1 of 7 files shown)


NeronCrypto (commented):

aaronjmars:
Hey @0xShak — I see you closed both of these in the meantime. I did a deep review of the pair today and the implementation itself is largely solid: the sidecar wiring here cleanly mirrors the existing gateways, and the buyer-side payment hardening in #415 (cap-before-sign, single-payment guard, inert-by-default) is genuinely well done.

That said, routing the fleet's full inference stream (and payments) through a gateway is a high-trust integration, and I couldn't verify BEAMR publicly — askbeamr.com is only a couple of weeks old, there's no public source or docs yet, and your relationship to the project isn't stated in either PR.

If you're behind BEAMR (or can connect me with whoever is), ping me on Telegram: **@aaronjmars** — happy to chat, and if the project checks out I'm open to landing both of these.

PR #415, opened 2026-06-10
feat(skill): beamr-route — pay-per-call inference over x402 with onchain receipt

0xShak:
## What

Adds a new skill, **`beamr-route`**, that sends a prompt through a [BEAMR](https://askbeamr.com) inference gateway, pays for that single call in **USDC on Base over x402**, and reports the answer **plus the onchain settlement receipt** (tx hash + basescan link).

BEAMR is an OpenAI-compatible **router**: it classifies the request, routes it to the cheapest capable provider, runs the completion, and settles the exact cost per call via the x402 `exact` scheme. Where the [companion gateway PR](https://github.com/aaronjmars/aeon/pull/414) routes *all* of Aeon's traffic through BEAMR silently, this skill is an on-demand task whose output is a **verifiable onchain artifact** — a natural fit alongside `x402-monitor` and Aeon's x402 tracking.

## Files

- **`skills/beamr-route/SKILL.md`** — standard skill flow: input guard → graceful skip when unconfigured → on-demand `x402-fetch` install → pay+infer → notify (answer + model + basescan link) → log spend to memory.
- **`skills/beamr-route/scripts/beamr-pay.mjs`** — buyer-side x402 client (`createSigner` → `wrapFetchWithPayment`), emits one JSON line. Per-call spend is capped by `BEAMR_MAX_PAY_USDC` (default `0.05`) — an offer above the cap throws rather than overpaying.
- **`.github/workflows/aeon.yml`** — inject `BEAMR_GATEWAY_URL` / `BEAMR_PAYER_KEY` (+ `BEAMR_NETWORK` / `BEAMR_MAX_PAY_USDC` vars) into the run step, matching how every other skill key is wired.
- **dashboard** — surface the two secrets under **Skill Keys**.

## Safety / opt-in

The skill is **inert until configured**: with `BEAMR_GATEWAY_URL` / `BEAMR_PAYER_KEY` unset it skips cleanly (`exit 0`), so it never fails a run or moves funds by default. The payer key is a wallet secret — the SKILL.md and the dashboard descriptor both tell operators to **fund a dedicated, low-balance wallet** and keep `BEAMR_MAX_PAY_USDC` small. This mirrors the existing pattern for fund-moving secrets (e.g. `BANKR_API_KEY`).

## Relationship to #414

Independent and complementary — #414 adds BEAMR as a *gateway* (route everything), this adds a *skill* (ask once, get a receipt). Either can merge alone; neither depends on the other.

## Validation

- `node --check skills/beamr-route/scripts/beamr-pay.mjs` ✅
- YAML parse of `aeon.yml` ✅
- Frontmatter matches the `create-skill` schema (`name`/`description`/`var`/`tags`/`requires`).

### DIFF
--- a/.github/workflows/aeon.yml
+++ b/.github/workflows/aeon.yml
@@ -292,6 +292,10 @@ jobs:
           DEVTO_API_KEY: ${{ secrets.DEVTO_API_KEY }}
           NEYNAR_API_KEY: ${{ secrets.NEYNAR_API_KEY }}
           NEYNAR_SIGNER_UUID: ${{ secrets.NEYNAR_SIGNER_UUID }}
+          BEAMR_GATEWAY_URL: ${{ secrets.BEAMR_GATEWAY_URL }}
+          BEAMR_PAYER_KEY: ${{ secrets.BEAMR_PAYER_KEY }}
+          BEAMR_NETWORK: ${{ vars.BEAMR_NETWORK }}
+          BEAMR_MAX_PAY_USDC: ${{ vars.BEAMR_MAX_PAY_USDC }}
           JSONRENDER_ENABLED: ${{ vars.JSONRENDER_ENABLED || 'true' }}
           SKILL_VAR: ${{ inputs.var }}
           SKILL_NAME: ${{ steps.skill.outputs.name }}
--- a/apps/dashboard/app/api/secrets/route.ts
+++ b/apps/dashboard/app/api/secrets/route.ts
@@ -43,6 +43,8 @@ const BUILTIN_SECRETS: Omit<Secret, 'isSet'>[] = [
   { name: 'NEYNAR_SIGNER_UUID', group: 'Skill Keys', description: 'Neynar managed signer UUID — required to publish Farcaster casts' },
   { name: 'GH_GLOBAL', group: 'Skill Keys', description: 'GitHub PAT with cross-repo access — cross-repo skills & deploys' },
   { name: 'BASE_RPC_URL', group: 'Skill Keys', description: 'Custom Base RPC endpoint — onchain Base skills (fund-flow, honeypot-check, vigil-revoke). Optional: a public RPC is used by default; set a paid endpoint to lift rate limits.' },
+  { name: 'BEAMR_GATEWAY_URL', group: 'Skill Keys', description: 'BEAMR inference-gateway URL — the beamr-route skill pays per call over x402 (USDC on Base) and returns the answer + onchain receipt.' },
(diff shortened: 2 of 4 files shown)


.github/workflows/aeon.yml
```diff
@@ -292,6 +292,10 @@ jobs:
           DEVTO_API_KEY: ${{ secrets.DEVTO_API_KEY }}
           NEYNAR_API_KEY: ${{ secrets.NEYNAR_API_KEY }}
           NEYNAR_SIGNER_UUID: ${{ secrets.NEYNAR_SIGNER_UUID }}
+          BEAMR_GATEWAY_URL: ${{ secrets.BEAMR_GATEWAY_URL }}
```
NeronCrypto, line 295: [ISSUE] BEAMR skill secrets are injected here in `aeon.yml` but `messages.yml` is untouched — when the skill is triggered via Telegram (`/run beamr-route`), `BEAMR_GATEWAY_URL` and `BEAMR_PAYER_KEY` are absent and the skill hits the graceful-skip guard (exit 0 with notify), so it can never execute from message-triggered runs. Add the four BEAMR vars to the env block in `messages.yml` the same way PR #414 adds `BEAMR_BASE_URL`/`BEAMR_API_KEY`.

NeronCrypto (commented):
Missing `messages.yml` injection means the skill silently skips from Telegram-triggered runs.

- [ISSUE] .github/workflows/aeon.yml:295 — BEAMR skill secrets (BEAMR_GATEWAY_URL, BEAMR_PAYER_KEY, BEAMR_NETWORK, BEAMR_MAX_PAY_USDC) are only injected in one aeon.yml run job but absent from messages.yml; skill hits graceful-skip guard (exit 0) whenever triggered via Telegram, so it can never actually execute from message-triggered runs.

aaronjmars:
Same note as on #414 — if you want to pick this back up, ping me on Telegram: **@aaronjmars** so we can chat about BEAMR first.

For reference, the technical items from my review that would need addressing on reopen: `capabilities: [external_api, writes_external_host, onchain_writes, sends_notifications]` in the frontmatter, registration in `aeon.yml` (disabled, workflow_dispatch) + `./generate-skills-json`, a Sandbox note (does the node payment step run in the Actions sandbox, or does it need a shim like liquidpad-launch?), pinning the `x402-fetch` version, and labeling the settlement tx as gateway-reported unless it's verified against Base RPC. The spend-bounding itself was the strongest I've reviewed in this repo — nice work on that part.

PR #510, opened 2026-06-19
Add LENS skill pack

Tholynceus:
## What

Adds the LENS skill pack to the community skill packs registry.

## Why

LENS reads the on-chain deployer wallet behind any X profile and returns a CLEAR/CAUTION/STOP rug-risk verdict via the public LENS API. Resolves the dev behind a handle, flags rug/scam token devs on Base / Bankrbot. No key, public data only. Pack repo: https://github.com/Tholynceus/lens-skill-pack

## Type of change

- [x] Community skill pack listing

---

### Community skill pack listing

- [x] Pack repo is public with a clear license (MIT / Apache-2.0)
- [x] Pack has a `skills-pack.json` manifest at its root and a `SKILL.md` per skill
- [x] Write / onchain / bet skills are `default_enabled: false`
- [x] This PR adds **both** a README table row and a matching `skill-packs.json` entry
- [x] No monkey-patching of Aeon internals; no private or auth-walled endpoints required to run

bios lie, chains don't

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Installed skills land in `skills/` and are added to `aeon.yml` disabled - flip `
 
 ## Why "the most autonomous"?
 
-Most agent tools put you in the driver's seat - approve this tool call, review this diff, confirm this action. Aeon is built for the work you want *done* while you're not there: briefings, market monitoring, PR reviews, research digests, security scans.
+Most agent tools put you in the driver's seat - approve this tool call, review this diff, confirm this action. Aeon is built for the work you want *done* while you're not there: briefings, market monitoring, PR reviews, research digests.
 
 |  | Aeon | Claude Code | Hermes | OpenClaw |
 |--|------|------------|--------|---------|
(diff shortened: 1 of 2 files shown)


aaronjmars:
Thanks for the contribution, @Tholynceus — I reviewed the pack end to end (read every file, hit the endpoint live, checked the SKILL.md for safety) and it looks clean and genuinely read-only. 👍

Two small changes before I merge:

1. **Pin a stable endpoint domain.** The skill calls `https://lens-liard.vercel.app/api/lookup` — that's an auto-generated Vercel hostname, so the skill would silently break if the project ever redeploys under a different project name. Could you point it at a stable domain (e.g. `api.lnsx.io`) instead?

2. **Mirror the capabilities per-skill.** The registry entry declares `capabilities: ["read_only", "external_api"]`, but the pack's own `skills-pack.json` / `SKILL.md` frontmatter don't declare any. Per our `docs/CAPABILITIES.md`, the registry value should be the union of each skill's per-skill capabilities — so please add `["read_only", "external_api"]` to the LENS skill in the pack manifest so they stay in sync.

Once those two are in, I'll merge. Thanks again!

aaronjmars:
Thanks for this, @Tholynceus 🙏 The registry listing itself is clean — well-formed `skill-packs.json` entry, matching README row, capabilities in taxonomy, and the skill code is safe (single read-only public GET, no secrets).

I'm holding the merge on one blocker that lives in the **linked pack repo** (`Tholynceus/lens-skill-pack`), not in this PR: the pack can't be installed as published.

In `skills-pack.json`, the skill is declared with `name` where the installer requires `slug`:

- `install-skill-pack` reads `.skills[$i].slug` and aborts with `"Manifest skill #N has no slug — aborting."` (exit 1) when it's missing. There's no fallback to `name`.
- `path` is set to the **file** (`skills/lens-scan/SKILL.md`), but the installer treats `path` as a **directory** and appends `/SKILL.md` itself — so it resolves to `skills/lens-scan/SKILL.md/SKILL.md` and won't be found.

Both are one-line fixes in your pack's `skills-pack.json`:

```diff
 "skills": [
   {
-    "name": "lens-scan",
-    "path": "skills/lens-scan/SKILL.md"
+    "slug": "lens-scan",
+    "path": "skills/lens-scan"
   }
 ]
```

For reference, every existing community pack (e.g. `careful-finance-aeon-skill-pack`, `hunch`, `aeon-skill-pack-polymarket`) uses `slug` + a directory `path`.

Could you apply that and re-run `./scripts/validate-pack.sh` until it exits 0? That's the publishing checklist step that catches this. Once the pack installs cleanly I'm happy to merge — the listing is otherwise good to go. Thanks!

PR #558, opened 2026-06-27
vigil skill: refresh 9 → 17 tools

vigilcodes:
The `vigil` skill listed **9** read-only tools; VIGIL now ships **17**. This refreshes the skill to match what's live at mcp.vigil.codes.

## Added tools
- `vigil_consensus` — 6-source verdict (false-positive guard)
- `vigil_liquidity_lock` — locked/burned/unlocked/unknown (never 'safe' on missing data)
- `vigil_check_tax` — punishing or owner-modifiable buy/sell/transfer tax ('0% now, 99% later')
- `vigil_check_ownership` — mint / pause / blacklist / reclaim-ownership / modify-balances / selfdestruct
- `vigil_detect_clone` — copy-paste scam clones by bytecode fingerprint
- `vigil_check_scam` — community scam reports
- `vigil_sentinel_status` — autonomous watchlist config
- note for `vigil_simulate_approval` (spender+token pair)

## Changes
- `skills/vigil/SKILL.md`: tool list 9→17, frontmatter description, new call steps, x402 pay-per-call note
- `aeon.yml`: one-line description refreshed
- regenerated `skills.json` + `packs.json` per CONTRIBUTING (ci-skills-json / ci-packs-json)

Read-only, autonomous-safe. The write action stays in the separate `vigil-revoke` skill.


aaronjmars:
Thanks for the refresh @vigilcodes — clean diff, green CI, and it follows the existing vigil-skill conventions nicely. One thing to fix before merge:

The skill advertises **17 tools** but only **16 are actually wired**. Tool #17 "Sentinel Status" (`vigil_sentinel_status`) is listed at `SKILL.md:30` but has no `vigil_call` step anywhere, so an agent running the skill can't actually invoke it.

Could you either:
- add the missing `vigil_call vigil_sentinel_status` step (preferred, if the endpoint is live), or
- drop it from the list and change the count/description to 16,

so the advertised capability matches what the skill can actually run?

Minor (non-blocking): the x402 note about premium tools settling "a few cents of USDC per call" is aspirational — the `vigil_call` helper does a plain curl with no payment handling and treats any non-200 as `VIGIL_HTTP_ERROR`. Worth a quick wording tweak so it doesn't imply a capability that isn't implemented. Happy to merge once the tool count is squared away.

PR #559, opened 2026-06-27
feat(phylax-audit): hosted-engine fast path for canonical verdicts

usephylax:
## What

Adds a **hosted-engine fast path** to the `phylax-audit` skill. Before doing manual line-by-line work, the skill now calls the canonical Phylax API:

```
GET https://usephylax.com/api/audit?skill=<ref>
```

If that returns a valid `ALLOW`/`WARN`/`DENY` verdict, the skill uses it directly and can embed the badge (`/api/badge?skill=<ref>`). If the call fails, is rate-limited, or the sandbox blocks outbound network, it falls back to the existing inline 3-scanner method — same rules, same scoring.

## Why

`skill-triage` already runs a Phylax pre-screen before installs. Routing through the hosted engine makes that verdict:
- **Canonical** — identical to the npm package, CLI, and badge (same engine, same deterministic score).
- **Faster** — one call instead of re-deriving the scan in-prompt.
- **Resilient** — inline scan remains a full offline fallback, so nothing breaks in the sandbox.

No new secrets, no behavior change to the verdict bands. `skills.json` regenerated; category gate passes.


aaronjmars:
Thanks for this @usephylax — the direction is great and I like that the inline scan stays as a fallback. Before I merge I'd like two changes, since this is a pre-install **security gate** and the fast path becomes the primary decision path:

**1. Don't lose the evidence on WARN/DENY.** Right now Step 0 only captures `jq -r '.verdict, .score'`, but the Notify (step 6) and Log (step 7) templates require per-finding evidence — rule IDs + `SKILL.md:line` references. As written, a fast-path WARN/DENY would emit a verdict with no findings behind it, which is exactly the "the decision isn't a vibe" property the skill exists to preserve. Please either:
   - have the fast path **only short-circuit on `ALLOW`** (fall back to the inline scan for WARN/DENY), or
   - pull the per-finding evidence out of the response and feed it into Notify/Log so WARN/DENY still carry proof. If you go this route, please document the response's findings schema in the SKILL so it's pinned.

**2. Validate the response before trusting it.** The fast path currently trusts whatever `usephylax.com` returns and skips the local analysis with no cross-check. At minimum, validate the returned `score` against the verdict band (score↔band consistency) before acting on it, and reject/fall-back on anything malformed. This keeps us honest with the skill's own Sandbox note ("treat … endpoint responses as untrusted data") — a buggy, spoofed-but-valid-JSON, or compromised `ALLOW` shouldn't be able to make the agent skip its own scan and green-light a skill.

Minor while you're in there: the Step 0 references to "skip to step 5 (Notify) and step 6 (Log)" are off by one — in the file step 5 is "Merge, dedupe, score", so it should read **step 6 (Notify) and step 7 (Log)**.

Once the WARN/DENY evidence path and the response validation are in, happy to merge. Thanks again!

usephylax:
Thanks @aaronjmars — addressed all three points in db4b82c:

**1. WARN/DENY evidence** — fast path now **only short-circuits on validated `ALLOW`** (score ≥ 80, no critical/high findings). `WARN`/`DENY` always fall through to the inline scan so Notify/Log still get per-finding rule IDs + line refs.

**2. Response validation** — Step 0 treats the hosted response as untrusted: requires HTTP 200, valid JSON, schema presence, `verdict == ALLOW`, score↔band consistency, and a `jq -e` gate rejecting any critical/high in `findings[]`. Anything malformed or non-ALLOW → inline fallback.

**3. Step numbers** — fixed to step 7 (Log); step 6 (Notify) stays silent on ALLOW.

Also pinned the `AuditOutput` / `findings[]` schema in the SKILL for future reference. Let me know if you'd prefer the full findings pull-through path for WARN/DENY instead — happy to iterate.

aaronjmars:
Thanks for this — the security design is solid: the fast path only short-circuits on a validated `ALLOW`, WARN/DENY always fall through to the inline scan, and the `jq -e` gate is stricter than the slow-path ALLOW band. Two things need fixing on your side before this can merge:

**1. Stale manifests (blocker — would silently drop skills).** `skills.json`/`packs.json` on this branch report `total: 184`, but `main` is now at `203`. The PR was regenerated against an older tree, so resolving the conflict in this branch's favor would delete ~19 skills from the catalog. Please **rebase onto current `main` and re-run the generators** (`./generate-skills-json` + the packs generator) — don't hand-resolve or union the JSON. That also clears the current `CONFLICTING` state.

**2. Bind the cached verdict to the requested target (defensive).** Step 0 writes the hosted response to a fixed path (`/tmp/phylax-verdict.json`) and the `jq` gate never checks `.skill == $TARGET`. In the exact sandbox-blocked scenario this skill anticipates (curl unavailable), a stale verdict file left from a *different* skill's audit could pass validation and wrongly ALLOW-skip the current skill's scan. Two small changes:
   - `rm -f /tmp/phylax-verdict.json` before the curl, and
   - add `.skill == $TARGET` (and fold the `HTTP_CODE == 200` check into the `jq -e` gate rather than leaving it in prose only).

Minor nit while you're in there: the Step 0 fallback prose says "skip steps 1–4" but should say "1–5" — step 5 (merge/dedupe/score) is what derives the verdict band.

Once it's rebased and regenerated, the repo's own CI (`ci-skills-json`/`ci-packs-json`) will run and confirm the catalog is consistent.

imancipate:
Thanks @usephylax. A maintainer review pass will follow; in the meantime no changes requested from this triage layer.

aaronjmars:
Merged — thanks @usephylax, the design here is solid: the fast path only short-circuits on a validated `ALLOW`, and WARN/DENY/malformed/network-blocked all fall through to the inline scan, so a hosted failure can never weaken the verdict.

I pushed one commit (`eb3f7c8`) directly to close out blocker #2 (bind the cached verdict to the target) rather than bounce it back to you:

- **`rm -f /tmp/phylax-verdict.json` before the curl** — so in the sandbox-blocked-network case, a stale verdict left from a *different* skill's audit can't be read as this skill's result.
- **Folded `HTTP_CODE == 200` into the `jq -e` gate** (it was prose-only), so a 429/non-200 body can't slip through.
- **Bound `.skill` to the target in the gate.** One thing worth flagging for your side: I checked the live engine and `.skill` is the *resolved source URL* (`skill: input.skill_source`), not the raw ref — so a literal `.skill == $TARGET` would always miss for the common `owner/repo` form and silently disable the fast path. I used `((.skill == $target) or (.skill | contains($target)))`, which matches both the raw-URL and `owner/repo` forms and safely falls back to the inline scan otherwise. Also updated the pinned schema/prose to document `.skill` as the resolved source, and fixed the "steps 1–4" → "1–5" fallback reference.

Verified the gate against the stale-different-target case (now correctly falls back) and a null `.skill` (no jq crash) before pushing; CI came back green. Appreciate the contribution.

usephylax:
Thanks for merging! 🙏

Friendly nudge on **GHSA-252c-qvvq-j2j9** — patch is live on main (`5fef9e4`), severity set to **high**, patched version filled. Advisory is still **draft** whenever you're ready to hit **Publish**.

No rush — just flagging so it doesn't sit in draft. Credit `usephylax` is fine to keep.

usephylax:
Thanks again for merging — Phylax hosted-engine path is live on our side.

Separate gentle nudge on [GHSA-252c-qvvq-j2j9](https://github.com/aaronjmars/aeon/security/advisories/GHSA-252c-qvvq-j2j9): security patch is on `main` (`5fef9e4`, F1–F5 + F5 env wiring). Advisory still draft/unpublished — ready whenever you want to publish. Happy to help verify anything pre-publish.

— @usephylax

PR #576, opened 2026-07-01
docs(claude-md): fix drift + add architecture overview

aaronjmars:
Audited `CLAUDE.md` against the real workflows and scripts (four parallel passes over Tools/Notifications, Memory/Chaining, Sandbox/Security, and gaps/hoisting), then fixed the drift, hoisted the conventions that were copy-pasted across dozens of skills, and added the requested architecture overview. Every claim below was verified against source (file:line in the notes).

Net: **129 → 108 lines**, ~2,107 → ~2,395 tokens per run. The accuracy fixes + trims net *lighter*; the increase is entirely the two requested new sections (`How Aeon works`, `Capability mode`).

## 🔴 Correctness fixes (doc was wrong)
- **Chain example didn't parse.** `chain-runner.yml:187-197` extracts `consume:` with a regex against the *same line* that matched `skill:`. The old example put `consume:` on its own line → **silently dropped**. Now single-line, matching `aeon.yml:472`.
- **Wrong sandbox root cause.** "sandbox blocks env var expansion" → it blocks **outbound network** (per `notify.sh:16`, `postprocess-replicate.sh:2`, every prefetch script).
- **`notify-jsonrender` isn't called by `./notify`.** `notify.sh:165-167` only drops a pending file (and only when `SKILL_NAME` is set); the conversion runs in a separate workflow step (`aeon.yml:730`).
- **Stale inbound priority.** Dropped "first message wins" — the poller now collects all channels (`messages.yml:498-591`).
- **Issue-tracker enum drift.** Old doc listed 12 categories / 4 statuses; the filer (`skill-health/SKILL.md:87`) offers 7 categories and only uses `open`/`resolved`. Replaced the enum block with a pointer to the schema owner (kills the drift instead of relocating it).
- **Missing 4th channel + flags.** Added SendGrid email (`notify.sh:150-162`) and `--title`/`--severity`/`--link` + the trace/severity suppression gotcha.
- **Rules.** "one paragraph max" contradicted 45 skills using `./notify -f`; "rm -rf" reworded to note the allowlist structurally forbids it.

## 🟢 Additions (hoisted universal conventions)
- **Notify only on signal** (hand-rolled in ~53 skills), **dedup vs recent logs** before notifying (~84 skills), **parseable `### <skill>` log format** (the health loop parses it).
- **`## Capability mode`** — `read-only` strips `Write`/`Edit`/`Bash(git:*)`/`Bash(gh:*)`/python from `--allowedTools` (`skill_mode.sh:13`); prevents undiagnosable permission errors for the 25 skills that declare it.
- **`## How Aeon works`** — the requested end-to-end runtime overview.

## 🟡 Trims
- Compressed `Skill Chaining` (also removes the broken example), **cut** the local-only MCP section (self-admittedly irrelevant on the Actions path; referenced a `@json-render/mcp` package that isn't a dependency), compressed the Notifications channel table.

## Note / deviation
Rather than copy the issue enums into `docs/CORE.md` (which would just relocate the drift), I pointed CLAUDE.md at the actual schema owner — `skills/skill-health/SKILL.md`. `docs/CORE.md` already documents the loop, so no doc change was needed there.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,6 +2,19 @@
 
 You are Aeon, an autonomous agent running on GitHub Actions via Claude Code.
 
+## How Aeon works
+
+Aeon is a fork-and-configure agent framework. The operator enables **skills** (self-contained `SKILL.md` capabilities under `skills/`) and schedules them in `aeon.yml`. Each run is a fresh, headless Claude Code invocation — there is no long-lived process and nothing persists between runs except the `memory/` directory and the git repo itself.
+
+One skill run, end to end:
+1. **Dispatch** — a schedule or a manual **Run now** fires a single skill. Chains dispatch their steps through `chain-runner.yml`.
+2. **Resolve** — the workflow picks the model and the capability mode (`read-only` vs `write`, from the skill's frontmatter), resolves `.mcp.json`, and runs any `scripts/prefetch-*.sh` with full env access (the sandbox blocks that network later).
+3. **Run** — it launches `claude -p "run skill X"`. This file (`CLAUDE.md`) and `STRATEGY.md` auto-load as your standing instructions; the prompt points you at `skills/X/SKILL.md`, which you read and execute.
+4. **Act** — read memory, fetch/compute, write files or open a PR (write mode only), and report via `./notify`.
+5. **After** — on success the workflow runs `scripts/postprocess-*.sh` (deferred network side-effects), converts feed output via `./notify-jsonrender`, and reverts stray writes from read-only skills. You append a log to `memory/logs/`.
+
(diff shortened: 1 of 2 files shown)


aaronjmars:
**Follow-up (2nd commit `0e5e3f1`): chased down the removed `MCP Servers (local mode only)` command.**

The old `npx @json-render/mcp --catalog apps/dashboard/lib/catalog.ts` never worked. Findings:
- `@json-render/mcp` **does** exist on npm (v0.19.0) — but per the [official docs](https://json-render.dev/docs/api/mcp) it's a **library** (`createMcpApp`), **not** a CLI with a `--catalog` flag. Canonical usage is your own `server.ts` run via `npx tsx server.ts --stdio`.
- The repo has **no** such server (`apps/mcp-server` exposes Aeon *skills* as tools, unrelated), **no** `createMcpApp` usage, and **no** json-render entry in `.mcp.json.example`. So local json-render MCP was never wired — the command was fictional.
- `./aeon` is dashboard-only; it never runs skills/Claude, so 'running ./aeon locally uses the MCP tool' was doubly wrong.

This commit just corrects the two docs (CLAUDE.md + dashboard README) to stop claiming a working local json-render MCP. Actually *building* the server (createMcpApp `server.ts` + `.mcp.json.example` entry + adding the dep) is a separate feature, left as a follow-up.

PR #630, opened 2026-07-04
fix: reject duplicate slugs in skill-pack manifest

BBridgeers:
## Problem

`install-skill-pack` and `validate-pack.sh` both iterated `skills[]` without checking for repeated slug declarations. A manifest that accidentally listed the same slug twice would silently overwrite the first skill's install — the author gets no signal until the wrong skill config shows up at runtime.

## Fix

Add a space-delimited seen-slug check to both scripts:
- `install-skill-pack`: exits 1 on the first duplicate (fail-fast at install time)
- `validate-pack.sh`: reports it as an ERROR and keeps going so the author sees all problems in one pass

Uses a space-delimited string pattern (`case " $seen_slugs " in`) instead of associative arrays for macOS bash 3 compatibility — matching the existing convention noted at line 304 of `install-skill-pack`.

## Tests

New `scripts/tests/test_validate_pack.sh` covers three cases:
1. Duplicate slug → ERROR, exit 1
2. Clean pack (unique slugs) → passes, exit 0
3. Triple duplicate → ERROR, reports the slug

All tests pass:
```
ok   - duplicate slug detected with correct message
ok   - clean pack passes validation
ok   - triple duplicate detected

All validate-pack tests passed.
```

### DIFF
--- a/bin/install-skill-pack
+++ b/bin/install-skill-pack
@@ -345,6 +345,7 @@ if [[ -f "$MANIFEST_PATH" ]]; then
     exit 1
   fi
 
+  seen_slugs=""
   for i in $(seq 0 $((skill_count - 1))); do
     slug=$(jq -r ".skills[$i].slug // empty" "$MANIFEST_PATH")
     if [[ -z "$slug" ]]; then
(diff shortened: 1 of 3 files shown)


imancipate:
Thanks @BBridgeers. A maintainer review pass will follow; in the meantime no changes requested from this triage layer.

PR #693, opened 2026-07-08
fix(vuln-scanner): make the scan arm functional (allowlist grant + tool-staging step)

aaronjmars:
## Makes vuln-scanner's scan arm actually functional

A live-test (on the `aeon-sb` fork) surfaced that vuln-scanner's scan arm has been **silently non-functional** — every run fell back to manual code review. Two root causes, both fixed here, both cross-confirmed against the operator's working **`aeon-vuln`** instance (which arrived at the identical solution).

### 1. Allowlist grant — `scripts/skill_mode.sh`
The scanner bare-names (`semgrep` / `osv-scanner` / `trufflehog` / `slither`) were never in the write-tier allowlist, so `claude -p` **denied** them (`"requires approval"`) — and the model logged that denial as *"Blocked by sandbox"* (the myth, resurfacing at runtime). Adds them to `WRITE_TOOLS` + the `GROK_WRITE_BASH` mirror. Read-only tier untouched.

✅ **Validated on aeon-sb** (scanning `aaronjmars/minitor`): semgrep flipped from `fail — "Blocked by sandbox"` → **`semgrep=ok`**.

### 2. Tool staging — `scripts/stage-vuln-scanner.sh` + workflow step
The grant alone gets **semgrep** working (pip installs it onto the global PATH) but not `osv-scanner` / `trufflehog` — Go binaries staged into `/tmp/bin`, which don't resolve by bare name because Claude runs each Bash tool call in a **fresh shell** (the skill's `export PATH=/tmp/bin:$PATH` doesn't survive to the scan call). Only a workflow step can append `/tmp/bin` to `$GITHUB_PATH` and make it persist.

Adds a **purpose-built, vuln-scanner-only** step (`if: skill == 'vuln-scanner'`) that installs the four scanners and writes `$GITHUB_PATH`. This is binary **staging**, **not** a network workaround (there is no network sandbox) — so the generic `prefetch-*.sh` convention stays removed: it's one conditional step + one self-guarded script. Same mechanism proven in `aeon-vuln`.

⏳ **Validating on aeon-sb** with the full fix (allowlist + staging) → expecting `trufflehog=ok` + `osv=ok`; result will be posted as a comment.

## Files
- `scripts/skill_mode.sh` — write-tier grant (+ grok mirror)
- `scripts/stage-vuln-scanner.sh` — **new**; installs the 4 scanners + writes `$GITHUB_PATH`; self-guards to vuln-scanner; best-effort (never fails the run)
- `.github/workflows/aeon.yml` — conditional "Stage vuln-scanner tools" step before Run

No change to `skills/vuln-scanner/SKILL.md` — its in-run install stays as a fallback for runs without the step, and its prose is cleaned separately in #692. `test_skill_mode.sh` passes 22/22; `actionlint` clean.

**Capability note:** this widens the write-tier allowlist to four read-only static-analysis tools. Flagged for conscious review.

### DIFF
--- a/.github/workflows/aeon.yml
+++ b/.github/workflows/aeon.yml
@@ -272,6 +272,17 @@ jobs:
             echo "::warning::could not materialize cron-state from the Issue; readers use the committed file"
           fi
 
+      - name: Stage vuln-scanner tools
+        # Purpose-built binary staging for vuln-scanner's scan arm — NOT a network
+        # workaround (there is no network sandbox). Installs semgrep/trufflehog/
+        # osv-scanner/slither and appends /tmp/bin to the PATH file so the scanner
+        # bare-names resolve in the later claude -p step and its per-call fresh shells
+        # (an in-run `export PATH` doesn't survive between Claude's Bash calls).
+        # Execution is separately gated by the write-tier grant in scripts/skill_mode.sh.
+        # Best-effort: the script self-guards to vuln-scanner and never fails the run.
+        if: steps.work.outputs.mode != '' && steps.skill.outputs.name == 'vuln-scanner'
+        run: bash scripts/stage-vuln-scanner.sh "${{ steps.skill.outputs.name }}"
+
       - name: Run
         id: run
         if: steps.work.outputs.mode != ''
--- a/scripts/skill_mode.sh
+++ b/scripts/skill_mode.sh
@@ -40,6 +40,13 @@ BASE_TOOLS="$BASE_TOOLS,Bash(head:*),Bash(tail:*),Bash(wc:*),Bash(sort:*),Bash(g
 # Write tier additionally gets repo-mutation tools + python (an interpreter is itself
 # a write vector, so it stays out of the read-only base; skills' python helpers run here).
 WRITE_TOOLS="Write,Edit,Bash(gh:*),Bash(git:*),Bash(python3:*),Bash(python:*)"
+# Security-scanner bare-names for vuln-scanner (Arm A). The skill stages these in-run
+# (`python3 -m pip install` for semgrep/slither, `curl -o … && chmod +x` for the Go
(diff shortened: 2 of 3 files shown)


aaronjmars:
## ✅ Full-fix validation (aeon-sb)

Re-ran vuln-scanner on `aeon-sb` (branch patched with this PR's allowlist grant **+** staging step) against `aaronjmars/polymarket-tg-bot`.

**"Stage vuln-scanner tools" step — completed/success, all four scanners installed:**
```
semgrep 1.168.0   trufflehog 3.95.8 → /tmp/bin   osv-scanner → /tmp/bin   slither 0.11.5
manifest: semgrep=installed  trufflehog=installed  osv-scanner=installed  slither=installed
```

**Scan result (Arm A stdout — authoritative):**
```
Scanner results: semgrep=ok (2 low-severity CI supply-chain warnings), osv=ok (0 CVEs in 44 packages), trufflehog=skip
Candidates reviewed: ~6.  Confirmed: 2.
```

Before → after, same skill, same fork:

| tool | before | after (grant + staging) |
|---|---|---|
| **semgrep** | `fail` — "Blocked by sandbox" (allowlist denial) | **`ok`** |
| **osv-scanner** | `fail` — env restriction (never on persistent PATH) | **`ok` — scanned 44 packages** |
| **trufflehog** | `fail` — env restriction (not installed) | installed + on PATH; `skip` = no verified secrets this run |

`osv=ok` across 44 packages is the key confirmation — the Go binary now resolves by bare name across Claude's fresh per-call shells, which is exactly what appending `/tmp/bin` to `$GITHUB_PATH` in the staging step fixes. The two earlier runs (openroutercli, minitor) had all scanners failing and fell back to manual review; this run's scanners actually executed.

_(The run's overall `failure` conclusion is the unrelated post-push `git checkout main` step on the single-branch fork checkout — the skill and scanners ran fine.)_

PR #694, opened 2026-07-08
fix(workflow): don't fail Commit/cron steps on a non-main-ref dispatch

aaronjmars:
## What

The `Commit results` and `Update cron state` steps both assume `main` is a **local branch**. That's only guaranteed when the run *started* on main — the normal scheduled / Run-now path. A run dispatched against a **non-main ref** (`gh workflow run … --ref <branch>`) gets a single-branch checkout with **no local main**, and both steps break:

| Step | Failure | Cause |
|---|---|---|
| **Commit results** | exit 1 — `error: pathspec 'main' did not match` | unguarded `git checkout main` under `set -e` |
| **Update cron state** | exit 128 — `Committing is not possible because you have unmerged files` | guarded `checkout main` no-ops → `git pull --rebase origin main` conflicts on the divergent branch → leaves unmerged files → later `git commit` dies |

Both surfaced on the `aeon-sb` fork validation runs (dispatched on a test branch): every vuln-scanner test-run reported `conclusion=failure` even though the skill and scanners ran fine. The model then logs that as a spurious skill failure.

**Scope, honestly:** the normal **scheduled / Run-now-on-main** path never hits this — main is always a local branch there, so behaviour is unchanged. This bites (a) operator fork/branch test-runs, and (b) any real deployment that dispatches a skill on a non-main ref (repair / chain / manual-branch runs), where the false `failure` conclusion feeds `skill-health` scoring.

## Fix

Guard both spots on `git show-ref --verify --quiet refs/heads/main`:

- **main is a local branch** → identical behaviour to today (switch back, rebase, commit).
- **main absent** (non-main dispatch) → `Commit results` pushes the feature branch and stops (`exit 0`); `Update cron state` skips the commit (it's already non-fatal by design).

No `${{ }}` added to any `run:` block. `actionlint -shellcheck= -pyflakes=` clean. Guard verified locally against both branch states (main-present → true, single-branch → false).

## Files
- `.github/workflows/aeon.yml` — `Commit results` + `Update cron state` steps

### DIFF
--- a/.github/workflows/aeon.yml
+++ b/.github/workflows/aeon.yml
@@ -1196,9 +1196,19 @@ jobs:
             git diff --staged --quiet || git commit -m "chore($LABEL): auto-commit $(date +%Y-%m-%d)"
             git push --force-with-lease -u origin "$CURRENT_BRANCH" || true
 
-            # Switch back to main and cherry-pick memory/log changes if any
-            git checkout main
-            git pull --rebase origin main || true
+            # Switch back to main to record memory/log changes. main is only
+            # guaranteed present when the run STARTED on main (the normal
+            # scheduled / Run-now path). A run dispatched against a non-main ref
+            # gets a single-branch checkout with no local main, so `checkout main`
+            # would fail the step under `set -e` (and a blind rebase would
+            # conflict). The feature branch is already pushed above — just stop.
+            if git show-ref --verify --quiet refs/heads/main; then
+              git checkout main
+              git pull --rebase origin main || true
+            else
+              echo "main not present locally (non-main dispatch) — feature branch pushed, nothing to record on main"
+              exit 0
+            fi
           fi
 
           # Commit any remaining changes on main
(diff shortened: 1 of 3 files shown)


aaronjmars:
## ✅ Live-validated on aeon-sb

Dispatched `heartbeat` on `aeon-sb` against this fix branch (`--ref fix/commit-step-nonmain-dispatch`) — the exact non-main-ref scenario that reproduced both failures. [Run 28981334180](https://github.com/aaronjmars/aeon-sb/actions/runs/28981334180): **`conclusion: success`**, zero failed steps.

| step | before (non-main dispatch, no fix) | after (this PR) |
|---|---|---|
| overall | `failure` | **`success`** |
| `Commit results` | exit 1 — `error: pathspec 'main' did not match` | **success** |
| `Update cron state` | exit 128 — `Committing is not possible because you have unmerged files` | **success** |

The `show-ref --verify refs/heads/main` guard took the main-absent branch in both steps (branch pushed, cron commit skipped) instead of dying. Scheduled-on-main behaviour is unchanged — the guard predicate is `true` there.

PR #797, opened 2026-07-30
fix(telegram): gate inbound on owner user, not just chat (public-chat button hijack)

aaronjmars:
## Problem

Telegram inbound was gated on **chat id only** (`TELEGRAM_CHAT_ID`), never on the tapping/sending **user**. In a 1:1 DM that is safe because `chat.id == user.id`. But when `TELEGRAM_CHAT_ID` is a **group/public chat**, the chat id is shared by every member, so **any member could command the bot** — dispatch skills, schedule weekly crons, snooze/mute — just by tapping a posted `🔁 Run again` / `📅 Schedule weekly` button.

Worse, Telegram delivers `callback_query` updates even when BotFather group-privacy mode is on (which otherwise hides plain group messages from the bot). So **buttons were effectively less protected than message input** — the exact asymmetry a public-chat operator hits.

Three gate points, all chat-only:
- poller callback — `messages.yml` (`CB_CHAT == TELEGRAM_CHAT_ID`)
- webhook `route` job re-check — `messages.yml` (`CHAT_ID == TELEGRAM_CHAT_ID`)
- Cloudflare Worker callback — `apps/webhook/src/worker.js` (`cb.message.chat.id === owner`)

## Fix

Add an **owner user** gate alongside the existing chat gate, on **both** inbound paths (poller + Worker + route job), for **callbacks and messages/replies**:

```
from.id == (TELEGRAM_ALLOWED_USER_ID || TELEGRAM_CHAT_ID)
```

New optional `TELEGRAM_ALLOWED_USER_ID` (numeric Telegram user id), mirroring the existing `DISCORD_ALLOWED_AUTHOR_ID` / `SLACK_ALLOWED_USER_ID` pattern.

- **DM setup (unset):** private `chat.id == user.id`, so the default `TELEGRAM_CHAT_ID` *is* the owner. Backward compatible, zero new config.
- **Group/public chat (unset):** group ids are negative, `from.id` is a positive member id — equality never holds. **Fail-closed: every tap/message ignored** until the operator sets `TELEGRAM_ALLOWED_USER_ID` to their own id.

The Worker still replies "This bot is private." to a stranger's private DM, but never to a non-owner member inside the owner's own group (avoids group noise).

## Files

- `.github/workflows/messages.yml` — poller callback + message gate, route-job `from_id` re-check, env wiring
- `apps/webhook/src/worker.js` — `ownerUid` gate on callback + message
- `apps/webhook/{wrangler.toml,.dev.vars.example,README.md}` — Worker config + docs
- `apps/dashboard/lib/secrets-catalog.ts` — new `TELEGRAM_ALLOWED_USER_ID` vault entry
- `docs/telegram-commands.md` (new **Owner gate** section), `docs/telegram-instant.md`, `.github/README.md`

## Verification

- `actionlint` (with shellcheck) clean — no new warnings in the edited blocks
- `node --check apps/webhook/src/worker.js` OK
- `scripts/tests/test_telegram_route.sh` — all 13 pass (router unchanged; gating is transport-layer)

## Operator action

Group/public-chat operators: set `TELEGRAM_ALLOWED_USER_ID` to your numeric user id (from [@userinfobot](https://t.me/userinfobot)) as a repo secret/var, and the Worker env for instant mode. DM operators: no action.


aaronjmars:
Added a second commit on the same theme: **auto-suppress inline buttons outside the owner's private DM** (`notify.sh`). Operator controls are only actionable by the owner, but a group message's buttons are visible+tappable by every member and can't be scoped to one viewer — so `notify` now omits inline buttons entirely when the target chat isn't a 1:1 DM (private chat id positive; group negative). The message still posts; opt back in with `TELEGRAM_GROUP_BUTTONS=1`. Complements the tap gate (defense in depth). Adds test coverage for suppress + opt-out.

aaronjmars:
**Revised the button-suppression signal.** Earlier commits keyed it to chat type (group vs DM); that was the wrong axis. It now couples to the actual thing asked for: **is `.github/workflows/messages.yml` enabled**. When that inbound workflow is disabled, no poller/router receives a tap or a `force_reply` answer, so `notify.sh` stops attaching inline buttons AND force_reply prompts (message body still posts). It resolves the workflow `state` via the GitHub Actions API (best-effort, cached once per run), suppressing only on a definitive `disabled_manually`/`disabled_inactivity`; `active` or unreachable fails open. Overrides: `TELEGRAM_FORCE_BUTTONS=1` (force attach), `AEON_MESSAGES_WF_STATE=<state>` (skip the API call). The old `TELEGRAM_GROUP_BUTTONS` knob is gone. The owner-user gate is unchanged and orthogonal.

PR #867, opened 2026-08-09
feat: add TaskMarket MCP server integration and delegate skill

prayItCompiles:
Add TaskMarket marketplace tools to the Aeon MCP server, exposing 10 taskmarket_* tools for browsing, creating, and submitting work on the Base onchain agent marketplace.

## Changes

- apps/mcp-server/src/taskmarket-tools.ts: MCP tool definitions and CLI wrappers with side-effect gate
- apps/mcp-server/src/taskmarket-tools.test.ts: 12 vitest tests
- apps/mcp-server/src/index.ts: Route taskmarket_* tools through MCP server
- skills/taskmarket-delegate/: Aeon skill for delegating work to TaskMarket
- aeon.yml: Register taskmarket-delegate as on-demand skill
- catalog/skills.json: Regenerate to include new skill + tools

All 12 tests pass. TypeScript compiles cleanly.


aaronjmars:
Closing in favor of #865, which lands the same `taskmarket-delegate` skill via a self-contained zero-dependency client that works in the Aeon runner. This PR's MCP-server approach shells out to a third-party `@lucid-agents/taskmarket` CLI that is never staged in the runner (so every tool errors at runtime), declares the skill `mode: read-only` while it performs on-chain spends (create/submit) and writes under memory/, is missing its eyebrowlock.json entry (fails ci-skill-integrity), and has no operator-authorization gate on `taskmarket_create_task` (which funds escrow). If we want the MCP-client tool surface later, we can revisit with the CLI staged, an eyebrowlock entry, the mode corrected, and a spend-authorization gate. Thanks for the contribution.

PR #870, opened 2026-08-11
feat: add read-only TaskMarket radar skill

dacode-dev:
## Summary

Add `taskmarket-radar`, a read-only Aeon skill that finds open TaskMarket delegation opportunities with public escrow evidence.

## Behavior

- Fetches the public TaskMarket API and filters by keyword, gross reward, expiry, open status, and escrow transaction hash.
- Reports exact task IDs, reward fields, deadlines, competition counts, escrow hashes, and public task URLs.
- Sends one notification only for new or materially changed opportunities and logs each run.
- Never creates, claims, bids, submits, signs, or spends; no credentials are required.
- Treats task descriptions as untrusted data and distinguishes platform opportunities from actual revenue.

## Validation

- `bash scripts/check-skill-categories.sh`
- `bin/generate-skills-json`
- `bin/generate-packs-json`
- `node scripts/validate-readme-catalog.mjs`
- `node scripts/validate-skill-packs.mjs`
- `git diff --check`

The new skill is included in the generated first-party catalog and Basics pack. TaskMarket was not present in the checked-out Aeon repository before this change.


aaronjmars:
Thanks for the contribution! I'm closing this in favor of #865 (taskmarket-delegate), which covers the same `GET /api/tasks` browse over the identical endpoint, plus create/submit, and is already all-green.

Two blockers here regardless: there's no `eyebrowlock.json` entry, so `ci-skill-integrity` hard-fails; and the skill declares `mode: read-only` but Step 7 appends to `memory/logs/${today}.md` - the read-only tier blocks `Write` and `>`/`>>`, so that ledger write silently no-ops every run.

Your escrow-tx-hash credibility filter is a genuinely nice idea, though - I'd welcome a small follow-up folding that into #865's browse ranking. Closing as not planned for now.

PR #872, opened 2026-08-15
fix(aeon-update): don't silently delete a currently-enabled skill retired upstream

Svector-anu:
## What

`aeon-update`'s 3-way classifier CLEAN-DELETEs any `skills/<name>/` path that upstream removed and the operator never modified locally (S6), but never checks whether `<name>` is still `enabled: true` in the operator's own `aeon.yml`. A PR from this skill could delete an actively-scheduled skill's directory while `aeon.yml` still references it, and nothing in the PR review surfaces that — the break only shows up later, when `validate-config.js`'s skill-refs check (or the skill's own next scheduled run) fails.

## Why

Hit this for real doing a manual upstream sync on a live instance: the recent restructure (#647) retired `verdikta-hunter`, `x402-monitor`, and `fear-divergence`, all three enabled and unmodified locally, which meant a plain `git merge upstream/main` silently deleted their directories while `aeon.yml` still had them scheduled. Had to manually cross-reference every clean-deleted skill against the live config to catch it before merging.

## Fix

Downgrade this specific case (removed upstream + unmodified locally + `enabled: true` in current `aeon.yml`) to a **CONFLICT** instead of a CLEAN-DELETE, with a new reason `enabled-skill-removed-upstream`. The skill directory stays untouched, and the PR body gets its own loud section for this case instead of it disappearing into "Applied cleanly":

```markdown
### ⚠️ Currently-enabled skills removed upstream
For each CONFLICT with reason `enabled-skill-removed-upstream` (S6):
- `verdikta-hunter` — enabled in your `aeon.yml`, deleted upstream in {commit(s)}. **Not deleted here** so nothing breaks. Pick one: keep it as a fork-only skill going forward (nothing else to do), or disable it in `aeon.yml` to match upstream's current default set.
```

Also handled the resolve-tracking side: `PENDING` conflicts normally resolve by comparing local vs. the upstream HEAD blob, but this reason has no HEAD blob (the path is deleted upstream) — added the correct resolution rule (resolved once the operator disables the skill, or upstream re-adds a path of that name).

## Scope

Docs-only change to `skills/aeon-update/SKILL.md` — no code, no schema, no frontmatter changes. `check-skill-categories.sh` passes; frontmatter untouched.

## Provenance

AI provider/model: anthropic / claude-sonnet-5
Client / agent tooling: Claude Code, direct interactive session (not a Slop/measured skill run — this is the aeonfun/aeon repo itself, not an elizaOS contribution-program project)

PR #874, opened 2026-08-16
fix(aeon-update): don't silently delete a currently-enabled skill retired upstream

Svector-anu:
Re-cut of #872 off clean `main` per @aaronjmars' feedback — this branch has only the one intended file, no instance divergence.

## What

The 3-way classifier CLEAN-DELETEs any `skills/<name>/` path upstream removed and the operator never modified locally — but never checked whether `<name>` is still `enabled: true` in the operator's own `aeon.yml`. A sync PR from this skill could delete an actively-scheduled skill's directory while `aeon.yml` still references it, with nothing in the PR review flagging it — the break only surfaces later, when `validate-config.js`'s skill-refs check (or the skill's own next scheduled run) fails.

Hit this for real doing a manual upstream sync: three skills this instance had enabled (`verdikta-hunter`, `x402-monitor`, `fear-divergence`) were retired in a large upstream restructure, unmodified locally, and would have been silently deleted by a plain merge.

## Fix

Downgrade this case to a **CONFLICT** (reason: `enabled-skill-removed-upstream`) instead of deleting — the directory stays, and the PR body surfaces it as its own loud section ("⚠️ Currently-enabled skills removed upstream") so the operator can't merge past it without noticing. Also updates the PENDING-reconciliation logic (S8), since this conflict class has no HEAD blob to diff against — it resolves once the operator disables the skill or upstream re-adds a path of that name.

## Diff

`skills/aeon-update/SKILL.md` only — 1 file, +8/−2.

Closes #872.

Claude-Session: https://claude.ai/code/session_01LDy5EqdS9TeKAzotH95bNY

### DIFF
--- a/skills/aeon-update/SKILL.md
+++ b/skills/aeon-update/SKILL.md
@@ -124,6 +124,8 @@ Decide `f`'s disposition from its `status` and a content 3-way (local-current vs
 
 Never CLEAN-DELETE a `skills/<name>/` directory whose `<name>` is not present in upstream's tree - fork-only skills are operator work and are structurally untouched (upstream's compare can only reference paths that exist upstream).
 
+Also never CLEAN-DELETE a `skills/<name>/` directory if `<name>` is currently `enabled: true` in the operator's `aeon.yml` (`grep -E "^  ${name}: *\{[^}]*enabled: true" aeon.yml`) - upstream retiring a skill the operator has actively scheduled is exactly the case `validate-config.js`'s skill-refs check exists to catch, but only *after* the PR is merged; nothing in the PR review itself would otherwise flag it. Downgrade this case to **CONFLICT** (reason: `enabled-skill-removed-upstream`) instead of deleting - the directory stays, and S9 surfaces it as its own loud PR-body section rather than folding it into "Applied cleanly" or the generic conflict list.
+
 ### S7. Apply CLEAN changes on a branch (sync mode only)
 
 If `MODE == report`, skip to S9. Otherwise:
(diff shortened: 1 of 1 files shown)


aaronjmars:
Thanks for this — the fix is correct and strictly better than the status quo: it only ever *downgrades* a CLEAN-DELETE to a CONFLICT when `enabled: true` is positively detected, so it can never regress the normal update path. Reading `aeon.yml` (not `skills.json`) is the right enabled-state source, the S8 no-HEAD-blob reconcile exception is sound, and the dedicated S9 section is a nice touch. Verified the detection against the real inline `aeon.yml` format — it matches, rejects `enabled: false`, and the `[^}]*` correctly stops before a trailing `# ... enabled: true` comment so it can't false-positive.

One gap worth closing before merge:

**`skills/aeon-update/SKILL.md` — the detection is inline-only.** The guard is:

```
grep -E "^  ${name}: *\{[^}]*enabled: true" aeon.yml
```

That only matches a single-line `{ ... }` entry. A skill configured in the block/multi-line form:

```yaml
verdikta-hunter:
  enabled: true
```

is **missed** and still falls through to CLEAN-DELETE — i.e. the exact bug this PR fixes still bites a multi-line-configured skill, silently. Notably `scripts/validate-config.js` (the check this PR text cites) already reads the skills block in *both* inline `{ }` and multi-line form, so as written the guard is narrower than the repo's own reader.

Canonical `aeon.yml` is 100% inline today, so real-world impact is low — but since the whole point is "don't silently delete an enabled skill," please either:

1. broaden the detection to also match the multi-line form (a small awk/python block that parses the `skills:` map the way `validate-config.js` does would be the robust fix), **or**
2. if you'd rather keep it a one-liner, explicitly state the inline-only assumption in the SKILL.md text so the limitation is documented rather than a silent miss.

Minor, non-blocking:
- The `^  ` (hard two-space indent) is fine for canon but brittle if an instance re-indents its `skills:` block.
- The S9 example uses the real name `verdikta-hunter` rather than a generic placeholder like the sibling `scripts/foo.sh` examples — cosmetic.

Broaden (1) and I'll merge. Thanks again!

PR #906, opened 2026-08-19
fix(scheduler): make reactive success_rate condition actually fire

aaronjmars:
## What

Reactive triggers advertise three `when:` conditions in `aeon.yml`:

```
consecutive_failures >= N
success_rate < X
last_status = <value>
```

The scheduler's inline evaluator only handled `consecutive_failures` and `last_status`. A trigger written as `when: "success_rate < 0.5"` matched no branch and was silently dropped -- no dispatch, no error. Dead config that reads as working.

## Fix

Extract single-condition evaluation into `scripts/reactive_when.sh` (testable), covering all three documented conditions. Adds the missing `success_rate` branch with `< <= > >=` and a `total_runs > 0` guard so a never-run skill (which sits at `0.0`) does not false-fire `< X`.

## Notes

- Behaviour for the two already-working conditions is unchanged; the new unit test pins it (`scripts/tests/test_reactive_when.sh`, 20 cases, wired into `ci-tests.yml`).
- Unparseable conditions now emit a `::warning::` once instead of being silently ignored.
- **Backwards compatible:** a repo with no reactive triggers, or only `consecutive_failures` / `last_status` triggers, dispatches byte-identically.

Found while auditing an INTEGRATION.md porting plan against the repo; this is a pre-existing doc/impl drift the plan didn't catch, separate from the ported features. No `CLAUDE.md` / `AGENTS.md` change needed.

## Test

```
bash scripts/tests/test_reactive_when.sh   # 20 passed, 0 failed
node scripts/validate-config.js            # CLEAN
```

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -87,5 +87,7 @@ jobs:
         run: bash scripts/tests/test_cron_due.sh
       - name: circuit-breaker scheduler decision tests
         run: bash scripts/tests/test_breaker.sh
+      - name: reactive-trigger when-condition tests
+        run: bash scripts/tests/test_reactive_when.sh
       - name: skill-scan calibration (fires on sinks, not benign syntax)
         run: bash scripts/tests/test_skill_scan_calibration.sh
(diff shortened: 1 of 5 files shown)


aaronjmars:
## Live validation

Fired the scheduler against a live reactive trigger on a private instance. The extracted evaluator matched and dispatched:

```
Reactive trigger: heartbeat (digest matched: last_status = success)
```

The two pre-existing conditions and the new `success_rate` branch all route through the same helper (`reactive_when.sh`), unit-tested (20 cases) + proven live.

PR #907, opened 2026-08-19
feat(validate-config): validate reactive-trigger references and conditions

aaronjmars:
## What

`validate-config.js` checks the reference surface a forker edits by hand -- the `skills:` block and every skill wired into a `chains:` pipeline -- so a dangling reference fails CI instead of surfacing when a cron fires. The `reactive:` block was the one dispatch path it never looked at.

A reactive trigger whose target or `on:` source names a pruned skill, or whose `when:` is mistyped, was a **silent no-op**: no dispatch, no error at config time, none at run time.

## Change

Adds a fourth check mirroring the skill-refs one for the reactive path:

- every reactive **target** skill resolves to `skills/<name>/SKILL.md`
- every non-wildcard **`on:` source** resolves too (`*` is allowed)
- every **`when:`** parses to a condition `scripts/reactive_when.sh` can evaluate: `consecutive_failures >= N` | `last_status = <v>` | `success_rate <op> X`

`collectReactiveRefs()` and `validateWhen()` are exported and unit-tested. The `when:` grammar is kept deliberately in lockstep with the runtime evaluator.

## Notes

- The live `aeon.yml` ships the reactive examples commented out -> parses to zero references -> stays `CLEAN`. Verified by a test.
- **Backwards compatible:** a repo defining no reactive triggers validates identically.
- This is the same validation the INTEGRATION.md plan wants for its proposed `on_failure` / chain `when:` keys, applied to the reactive mechanism that already ships. No `CLAUDE.md` / `AGENTS.md` change.

## Test

```
node --test scripts/validate-config.test.js   # 12 pass
node scripts/validate-config.js               # CLEAN (+ new PASS reactive line)
```

### DIFF
--- a/scripts/validate-config.js
+++ b/scripts/validate-config.js
@@ -20,6 +20,10 @@
 //      exists. This validates the full reference surface a forker edits by hand: the
 //      whole skills: block (enabled or not, inline `{ }` or multi-line) AND every
 //      skill wired into a chains: pipeline (parallel / skill / consume).
+//   4. Reactive-trigger integrity - every target skill and `on:` source in the
+//      reactive: block resolves to a real skill, and every `when:` parses to a
+//      condition scripts/reactive_when.sh can evaluate (same dangling-reference /
+//      dead-config class as 3, for the event-driven dispatch path).
 //
 // Output contract:
 //   - Exit 0 + only PASS lines  => CLEAN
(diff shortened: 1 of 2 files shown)


aaronjmars:
## Live validation

This validator is CI-only (green). Live-testing the reactive path also surfaced that the **runtime** scheduler parser was stricter than this validator (it required a quoted `on:` and skipped disabled source skills), so a trigger this validator accepts could still silently not fire. #910 fixes the runtime side (accepts bare `on:`, documents the enabled-source rule) so validator and runtime now agree.

PR #908, opened 2026-08-19
feat(audit): structured audit log of privileged actions

aaronjmars:
Implements **item 5** of the chassis INTEGRATION.md porting plan: a structured audit log.

## What

A framework that emails maintainers, pushes commits under a git identity, and holds a GitHub PAT should be able to answer "what did it do on the 14th" without reading Actions transcripts. (The existing `.jsonl` files are Claude Code *session transcripts* consumed by `run-actions-summary.sh` -- a different artifact.)

Adds an append-only JSONL record, one line per action that reaches outside the run, uploaded as the `audit-log-<run_id>` artifact (30-day retention):

```json
{"ts":"2026-08-18T14:03:11Z","run_id":"1234567890","skill":"digest","action":"notify","target":"channels=telegram delivered=true","exit":0,"secrets_used":["TELEGRAM_BOT_TOKEN"]}
```

## Redaction (the crux)

`secrets_used` is **names only**. Secret **values** never reach the file: `scripts/audit.sh` scrubs every credential-shaped env var out of each record in its **raw, base64, and url-encoded** forms before writing, failing toward over-redaction. The test is adversarial -- a secret with special chars embedded in a URL, a JSON body, an error message, and base64 -- and asserts none of the three encodings survive:

```
bash scripts/tests/test_audit.sh   # 11 passed (incl. all 4 adversarial encodings)
```

## Instrumented sinks

- `notify.sh` -- one record per notify call (channels + delivered)
- `secretcurl` -- records the call; **output and exit code still pass through unchanged** (byte-identical `exec curl` when auditing is off)
- the git push in `aeon.yml`'s commit step

## Done-when checklist (from the plan)

- ✅ each instrumented action appears once per run
- ✅ redaction test passes against all four encodings
- ✅ artifact uploads even when the run fails (`if: always()`)
- ✅ a run with no privileged action produces an **empty file, not a missing one**

## Notes

- The log lives **outside the workspace** (beside the notify queue) so a `read-only`-sandboxed skill can still append; a post-run step stages a copy for the artifact.
- **Backwards compatible:** the instrumentation is a no-op when `AEON_AUDIT_LOG` is unset (local runs / unit tests). Verified `test_notify.sh` still ALL PASS.
- **Known gap (out of scope):** `gh` PR/issue creation done from *inside* a skill's `claude -p` call is not wrapped -- that needs a `gh` shim, a bigger change. Workflow-level git push is covered; skill-issued `gh` writes are the follow-up.
- This is a prerequisite the plan flags for the eventual item 1b (infra-secret scoping). No `CLAUDE.md` / `AGENTS.md` change required (auditing is automatic, not a config key).

### DIFF
--- a/.github/workflows/aeon.yml
+++ b/.github/workflows/aeon.yml
@@ -694,6 +694,12 @@ jobs:
           : > "$AEON_AUDIT_LOG"
           echo "AEON_AUDIT_LOG=$AEON_AUDIT_LOG" >> "$GITHUB_ENV"
           export AEON_AUDIT_LOG
+          # Name the running skill in every audit record (audit.sh reads AEON_SKILL).
+          # Without this the notify/secretcurl subprocesses don't see $SKILL and each
+          # record's skill field falls back to "unknown".
+          AEON_SKILL="${{ steps.skill.outputs.name }}"
+          echo "AEON_SKILL=$AEON_SKILL" >> "$GITHUB_ENV"
+          export AEON_SKILL
 
           # notify / notify-jsonrender are the committed scripts/*.sh (testable; see scripts/tests/).
           cp scripts/notify.sh ./notify
--- a/.github/workflows/aeon.yml
+++ b/.github/workflows/aeon.yml
@@ -684,6 +684,17 @@ jobs:
           echo "AEON_PENDING_DIR=$AEON_PENDING_DIR" >> "$GITHUB_ENV"
           export AEON_PENDING_DIR
 
+          # Structured audit log (scripts/audit.sh): one append-only JSON line per
+          # privileged action that reaches outside the run (notify send, secretcurl,
+          # git push). Lives beside the notify queue OUTSIDE the workspace so a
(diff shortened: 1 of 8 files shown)


aaronjmars:
## Live validation (private instance, real secrets)

Ran `digest` end-to-end on a private Aeon instance. The `audit-log-<run_id>` artifact uploaded with real records, secret **values** redacted (names only):

```json
{"action":"secretcurl","target":"https://api.x.ai/v1/responses","exit":0,"secrets_used":["XAI_API_KEY"]}
{"action":"notify","target":"channels=telegram delivered=true","exit":0,"secrets_used":["TELEGRAM_BOT_TOKEN"]}
```

No token value appears anywhere in the artifact. Redaction holds live.

Fix pushed (`015606c`): records showed `"skill":"unknown"` because `AEON_SKILL` was not exported into the notify/secretcurl subprocesses. Now exported alongside `AEON_AUDIT_LOG`, so each record names the skill.

PR #910, opened 2026-08-19
feat(scheduler): reactive handler learns which skill tripped it (item 3)

aaronjmars:
Implements **item 3** of the chassis INTEGRATION.md plan (per-skill `on_failure` edges) -- reframed after reading the code.

> ⚠️ **Stacked on #906.** This branch includes #906's commit; review the second commit (`feat(scheduler): pass the tripping skill...`) for this PR's change. Rebase onto main after #906 merges.

## Why not a new `on_failure:` key

Aeon already has a per-skill failure edge: the `reactive:` mechanism dispatches a handler when a `when:` condition holds for a source skill (evaluated in `scheduler.yml`, present since #649). A single-failure handler is expressible **today**:

```yaml
reactive:
  notify-operator:
    trigger:
      - { on: digest, when: "consecutive_failures >= 1" }
```

Building a parallel `on_failure:` system would duplicate this. So this PR closes the one real gap instead and documents reactive as the blessed failure edge.

## The gap

A handler fired by `on: "*"` was never told **which** skill tripped it -- `skill-repair` had to re-derive the failed skill. Now the scheduler passes the matched source skill to the handler as its `var` (the plan's "handler receives var set to the failed skill's name"), unless the handler declares its own `var`, which wins.

## Also

`docs/CONFIGURATION.md` now documents reactive as the per-skill failure edge: the three conditions, the single-failure example, the source-as-var behaviour, and the **90-minute per-handler dedup** that bounds failure loops (the plan's "budget" -- already present; there is no `pkill` off-switch on Actions).

## Scope notes vs the plan's item 3

- **Same-run latency:** not built. Reactive fires on the next scheduler tick (≤5 min), not in the same run. The ≤5 min saving isn't worth a run-path change; the `heartbeat` -> `health:` issue path still backstops silent failures.
- **Depth-1 guard:** the 90-min dedup is the loop bound rather than a strict depth counter. Called out in the docs.
- **Backwards compatible:** a handler with an explicit var is unchanged; a repo with no reactive triggers is unaffected. Config validation of these refs ships in #907.

No `CLAUDE.md` / `AGENTS.md` change needed.

### DIFF
--- a/.github/workflows/scheduler.yml
+++ b/.github/workflows/scheduler.yml
@@ -405,6 +405,7 @@ jobs:
                 if [ "$WRC" -eq 0 ]; then
                   echo "Reactive trigger: $RSKILL (${CHECK_SKILL} matched: ${WHEN})"
                   TRIGGERED=true
+                  MATCHED_SOURCE="$CHECK_SKILL"   # the skill that tripped this trigger
                   break
                 elif [ "$WRC" -eq 2 ]; then
                   echo "::warning::reactive trigger for '$RSKILL' has an unparseable condition: ${WHEN}"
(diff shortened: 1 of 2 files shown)


aaronjmars:
## Live validation

Live scheduler output on a private instance:

```
Reactive trigger: heartbeat (digest matched: last_status = success)
Reactive heartbeat var <- source skill 'digest'
Dispatching: heartbeat (var: digest)
```

The handler received the tripping skill as its `var`.

Fix pushed (`ccb3726`): live-testing surfaced that the scheduler `on:` parser required quotes, so an unquoted `on: digest` (the form the docs use) silently dropped the trigger. Loosened to accept both quoted and bare, and documented that the source skill must be enabled for the trigger to evaluate.

PR #911, opened 2026-08-19
feat(chains): verdict routing on Haiku scores (when: on chain steps)

aaronjmars:
Implements **item 2** of the chassis INTEGRATION.md plan: verdict routing on Haiku scores.

## What

Aeon already scores every run 1-5 with Haiku (`memory/skill-health/<skill>.json`). Today that number is a chart someone looks at. This turns it into a control-flow edge -- a chain step runs only when a condition holds for the score of the skill it consumes:

```yaml
chains:
  editorial:
    schedule: "0 9 * * *"
    steps:
      - skill: review, consume: [draft]
      - skill: polish,  consume: [review], when: "score > 3"     # good -> polish
      - skill: rewrite, consume: [review], when: "score <= 3"    # weak -> rewrite
```

## Scorer timing (the plan's open question, answered)

**No scorer move needed.** The scorer runs as a *post-run step inside each skill's own `aeon.yml` run* (not end-of-chain) and commits `skill-health` before that run completes. `chain-runner.yml` already `git pull`s after each step (to fetch `output/.chains/`), so the consumed skill's score is on disk before the next step dispatches. Routing reads `.quality_score` directly.

## Rules (all enforced + tested)

- Operators `== != < > <= >=`. Equality compares as strings; ordering requires integers on **both** sides and **fails loudly** (no `"10" < "9"`).
- **A `when:` whose key is missing does not fire** -> the step is *skipped, not failed*. Forces every routing path to publish a score or fall through to an unconditional step.
- **`when: "score > abc"` is rejected at config time** by `validate-config.js` in CI, not at 3am.
- **Per-chain dispatch budget** (`max_dispatches:`, default 10). A chain that routes back on itself hard-fails with the step sequence -- the plan's billed-minutes replacement for chassis's `pkill`.

## Done-when checklist

- ✅ routes to different steps on a real Haiku score
- ✅ a `when:` on an unset key skips and logs the reason
- ✅ `when: "score > abc"` rejected at validation time (`validate-config.test.js`)
- ✅ a loop halts at the budget with a legible error
- ✅ chains with no `when:` behave identically to today

## Tests

```
bash scripts/tests/test_chain_when.sh   # 18 pass (incl. missing-key, non-int, invalid)
node --test scripts/validate-config.test.js
node scripts/validate-config.js         # CLEAN (+ new PASS chain-when line)
```

Independent of the other INTEGRATION.md PRs (touches `chain-runner.yml`, not the scheduler). `validate-config.js` also changes in #907; they add different checks and will need a trivial merge in `main()`/exports. No `CLAUDE.md` / `AGENTS.md` change needed. **Out of scope** (per the plan): compound/boolean expressions -- one key, one operator, one value.

### DIFF
--- a/.github/workflows/chain-runner.yml
+++ b/.github/workflows/chain-runner.yml
@@ -178,12 +178,20 @@ jobs:
           ON_ERROR=$(echo "$CHAIN_BLOCK" | grep -oP 'on_error:\s*\K[a-z-]+' || echo "fail-fast")
           echo "Error mode: $ON_ERROR"
 
+          # Total-dispatch budget for the whole chain. Chassis lets cycles run because
+          # the operator owns the box; Aeon runs on billed Actions minutes and a shared
+          # rate limit, so a chain that routes back on itself (via when:) gets a hard
+          # cap instead of a pkill. Default low; override per chain with max_dispatches:.
+          MAX_DISPATCHES=$(echo "$CHAIN_BLOCK" | grep -oP 'max_dispatches:\s*\K[0-9]+' || echo "10")
+          DISPATCH_COUNT=0
+          echo "Dispatch budget: $MAX_DISPATCHES"
+
           # Parse steps into an ordered array
           # Each step is either:
           #   - parallel: [a, b, c]
           #   - skill: name [, var: "val"] [, consume: [a, b]]
           STEP_INDEX=0
-          declare -A STEP_TYPE STEP_SKILLS STEP_VAR STEP_CONSUME
+          declare -A STEP_TYPE STEP_SKILLS STEP_VAR STEP_CONSUME STEP_WHEN
 
           while IFS= read -r line; do
             # Skip empty lines and non-step lines
@@ -204,6 +212,10 @@ jobs:
               # Extract optional consume
               if [[ "$line" =~ consume:\ *\[([^\]]+)\] ]]; then
                 STEP_CONSUME[$STEP_INDEX]=$(echo "${BASH_REMATCH[1]}" | tr -d ' ')
+              fi
+              # Extract optional when: conditional routing on the consumed skill's score
+              if [[ "$line" =~ when:\ *\"([^\"]+)\" ]]; then
+                STEP_WHEN[$STEP_INDEX]="${BASH_REMATCH[1]}"
               fi
               STEP_INDEX=$((STEP_INDEX + 1))
             fi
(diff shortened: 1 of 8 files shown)


aaronjmars:
## Live validation

Live chain run executing on a private instance: `digest` scored **4**, so the chain routes `when: "score >= 1"` (runs) vs `when: "score > 99"` (skips). Unit-tested (18 cases) + CI green regardless.

Fix pushed (`d8f105c`): proving this live required a real (uncommented) chain, which exposed that the bare `- skill: X, consume: [...]` step form used in the commented examples is **not valid YAML once uncommented** (the second colon trips the parser). The chain runner reads either form, so the examples now use flow-brace (`- { skill: X, ... }`) and the rule is documented.

_Will update this comment with the live routing log once the run completes._

aaronjmars:
### Live routing log (update)

Chain ran on the private instance. The routing **mechanism** is proven end to end:

```
Dispatch budget: 10
=== Step 1/3: single [digest] ===
=== Step 2/3: single [skill-health] ===
  Skip: when "score >= 1" not satisfied for 'digest' (verdict={})
=== Step 3/3: single [article] ===
  Skip: when "score > 99" not satisfied for 'digest' (verdict={})
```

The budget printed, each `when:` was evaluated against the consumed skill's verdict, both steps were **skipped (not failed)**, and each skip logged its reason - exactly the `missing key does not fire` semantic.

Both branches skipped rather than one running because the verdict was empty (`{}`): Step 1's internal `digest` run **timed out** in this instance (`Chain timed out waiting for runs`), so no score was committed for the later steps to read. That is a test-topology / instance issue (a `--ref` chain dispatching an internal run that did not finish), not a routing bug - and the empty-verdict path behaved correctly. The happy-path (a step running on a satisfied score, e.g. `score > 5` at `7`) is covered by the 18 unit tests in `test_chain_when.sh`.

aaronjmars:
### Clean happy-path (update 2)

Re-ran with the consumed skill's real Haiku score present in the read path. One branch runs, the other skips, both on the same score:

```
=== Step 1/2: single [skill-health] ===
  when "score >= 1" satisfied (verdict={"score":4})
=== Step 2/2: single [article] ===
  Skip: when "score > 99" not satisfied for 'digest' (verdict={"score":4})
```

`digest` scored **4**; `score >= 1` fired (step ran), `score > 99` did not (step skipped) - live, on real scoring data. #911 fully validated end to end.

PR #937, opened 2026-08-22
fix(webhook): dedupe Telegram updates by update_id before dispatch

Svector-anu:
## What

`apps/webhook/src/worker.js` is the untrusted-input entry point for Telegram's instant-delivery webhook. Its own `dispatch()` function carries the comment *"dedupe is by update_id / callback id"* — but nothing in the file ever actually checked it. Telegram redelivers an update (same `update_id`) whenever the webhook's previous response wasn't a prompt `200`: a slow dispatch, a transient 5xx from GitHub's `dispatches` API, a network blip on either side. Every redelivery re-dispatched — a slash command would run twice, a button tap would fire its action twice.

## Fix

A replay guard in `fetch()`, ahead of all classification/dispatch logic, keyed on `update.update_id` — present on every Telegram Update regardless of sub-type (`message`, `callback_query`, ...), so one cache covers every path through the handler uniformly. Only a genuine `200` gets cached; a failed dispatch (`502`) stays un-cached so Telegram's own retry can still get through and succeed — caching a failure would turn a transient error into a permanently dropped update. Extracted the existing classify+dispatch body into `handleUpdate()` so the guard wraps every path at one point instead of needing a check at each of the 4 `dispatch()` call sites.

**Storage: Cache API (`caches.default`), not KV/D1/DO.** Each operator deploys an isolated Worker with no provisioned storage (`wrangler.toml` has no bindings), consistent with this file's own "no shared infrastructure, no credential custody" design — Cache API needs no binding at all.

## Verification

**I live-verified the underlying mechanism actually works before writing the fix**, rather than trusting docs alone — Cache API support on `*.workers.dev` (the default deployment target in this project's own README quickstart) has genuinely conflicting reports between older community threads and current docs. I deployed a minimal probe Worker via `wrangler deploy --temporary` to a real `*.workers.dev` URL:

```
=== request 1 ===
MISS:wrote-new-entry
=== request 2 ===
HIT:written-1787422942907
```

A `cache.put()` from one request round-tripped as a `cache.match()` HIT on a later request, confirmed across multiple requests and separate process invocations. It's per-colo, not a global/distributed store (I also observed a MISS on what was evidently a different edge, after a HIT had already been proven on another) — so this closes the realistic single-region retry window rather than guaranteeing exactly-once delivery globally. That's documented as exactly that in the code comment, not oversold.

Added `apps/webhook/test/replay-guard.test.mjs` (`npm test`), which imports the **real** `src/worker.js` (not a hand-copied approximation) against a stubbed in-memory `caches` and a stubbed outbound `fetch` (`Request`/`Response`/`Headers` are real, native Node classes — the same Fetch API surface the Workers runtime implements):

```
$ npm test
ok   - first delivery of a new update_id dispatches and returns 200
ok   - redelivery of the same update_id is short-circuited, no second dispatch
ok   - a different update_id dispatches normally (dedupe is per-update_id, not global)
ok   - a failed dispatch is not cached, so Telegram's retry after failure still goes through
ok   - callback_query updates are deduped by the same update_id guard as messages
ok   - an update with no update_id is still processed normally (dedupe fails open)
ALL PASS
```

A small loader hook stubs the one `@microlabs/otel-cf-workers` import that needs the Workers runtime's `cloudflare:` scheme to resolve — that import is opt-in telemetry, never invoked when `OTEL_EXPORTER_OTLP_ENDPOINT` is unset, as in every test case here. Everything actually under test is the real, unmodified handler.

**Mutation-resistance**: ran the same test file against the actual pre-fix `worker.js` (fetched via `git show upstream/main:...`, not hand-copied) and confirmed it fails for exactly the predicted reason — a redelivered `update_id` dispatches twice instead of once.

## Attribution

gpt-5.6-sol (via Codex/AgentRouter) flagged the missing replay/idempotency check on this untrusted-input entry point in a round-2 audit of this repo, noting the gap between the code's own comment and its actual behavior. Claude Code (claude-sonnet-5) independently verified the finding, confirmed the Cache API mechanism actually works in the project's real default deployment target before committing to the approach, designed the fix, and built the test harness plus mutation-resistance verification.

### DIFF
--- a/apps/webhook/README.md
+++ b/apps/webhook/README.md
@@ -12,7 +12,8 @@ Worker** (`npx wrangler deploy`) after updating `src/worker.js` to pick up chang
 
 Each user deploys it into **their own** Cloudflare account. There's no shared
 relay and no credential custody — your bot token and GitHub PAT live only in your
-Worker's secrets.
+Worker's secrets. The replay guard also needs one small piece of Cloudflare
+infrastructure: a Workers KV namespace in that account.
 
 ## Deploy
 
@@ -20,15 +21,24 @@ Worker's secrets.
 
 > Forked Aeon? Change `aeonfun/aeon` in the button URL above to
 > `your-username/your-fork` so it deploys from your repo. (The button requires a
-> **public** source repo.)
+> **public** source repo.) The button is no longer deploy-and-done by itself:
+> create the required KV namespace and put its ID in your fork's
+> `apps/webhook/wrangler.toml` first.
 
 Or from a clone:
 
 ```bash
 cd apps/webhook
 npm install
+npx wrangler kv namespace create REPLAY_GUARD
+npx wrangler kv namespace create REPLAY_GUARD --preview  # for local wrangler dev
+# Paste the returned ids into [[kv_namespaces]] in wrangler.toml.
 npx wrangler deploy
 ```
+
+The KV namespace is required for replay protection on the normal `*.workers.dev`
+deployment target. Keep the production `id` and local `preview_id` separate; do
+not use a preview namespace for a deployed Worker.
 
 ## Configure
 
@@ -55,8 +65,9 @@ npx wrangler secret put GITHUB_TOKEN              # GitHub PAT (see scopes below
 | `GITHUB_REPO` | yes | `owner/repo` of your Aeon fork, e.g. `aeonfun/aeon` — not the worker repo the deploy button creates. |
 | `GITHUB_TOKEN` | yes | Fine-grained PAT scoped to your fork with **Contents: read/write** and **Actions: read/write**, or a classic token with `repo`. |
 
-To edit values later: Cloudflare dashboard → Workers & Pages → your worker →
-Settings → Variables and secrets.
+To edit secret values later: Cloudflare dashboard → Workers & Pages → your worker →
+Settings → Variables and secrets. To manage the replay namespace later, use
+Workers & Pages → KV or the `wrangler kv namespace` commands.
 
 ## Point Telegram at the Worker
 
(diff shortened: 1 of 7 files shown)


aaronjmars:
Thanks - the scoping here is right (the poller path already dedupes natively via Telegram's `getUpdates` offset-ack, so the webhook was the genuine gap), the failure semantics are good (cache only on 200, never on the 502 dispatch-failure path, so a transient GitHub error still lets Telegram retry), and the `handleUpdate()` refactor is clean. One blocker before it can merge, though:

**The dedup store is a documented no-op on the default deploy target.** The guard uses the Cloudflare Cache API (`caches.default`), but the README's deploy target is a `*.workers.dev` subdomain (`apps/webhook/README.md`, and the "Deploy to Cloudflare" button provisions a workers.dev Worker). Cloudflare documents that Cache API operations have no effect on `*.workers.dev` (and in the dashboard editor/Playground) - `cache.put()` stores nothing and `cache.match()` always misses. So on the documented default deployment this guard silently does nothing: every redelivery still re-dispatches, exactly the bug the PR sets out to fix. It only works behind a custom domain, which the README doesn't direct operators to set up. The code comments (worker.js ~line 84) and the test both assert the fix works, which will mislead maintainers into believing the webhook is deduped when for most operators it isn't. (The test's `Map`-backed `caches` stub validates control flow only, so it passes regardless of the workers.dev behavior.)

Refs: https://developers.cloudflare.com/workers/reference/how-the-cache-works/ and https://developers.cloudflare.com/workers/runtime-apis/cache/

Two ways forward:
- **(a) Use a workers.dev-reliable store** - KV with a short TTL (accepts adding a binding to `wrangler.toml`), or D1/Durable Object. This actually dedupes on the default target.
- **(b) Scope it honestly** - prominently document that instant-mode dedup requires a custom domain, warn on workers.dev, and soften the comments/test names so they don't over-claim.

Minor, either way: the `cache.match` (~line 84) vs `ctx.waitUntil(cache.put(...))` (~line 94) is a check-then-act TOCTOU (two concurrent redeliveries of the same `update_id` both miss and both dispatch), so "exactly once" isn't guaranteed even on a zone; and the `max-age=300` window means a redelivery after 5 min re-dispatches - worth a comment.

Svector-anu:
you're right, and i re-verified rather than argue: re-read cloudflare's own cache docs word for word, and they explicitly list "workers deployed to custom domains" (and pages functions) as having functional cache operations — a plain worker on workers.dev is conspicuously not in that list.

worth being honest about why i missed it: i did test this live before shipping, via `wrangler deploy --temporary` to a real workers.dev url, and genuinely saw a cache hit. that test was misleading — a "temporary preview account" deploy almost certainly provisions differently than the normal persistent workers.dev deploy real operators actually get. lesson taken.

switched to your option (a) — workers kv. pushed:
- `worker.js`: `env.REPLAY_GUARD.get()`/`.put()` instead of `caches.default`, same key/window
- `wrangler.toml`: `[[kv_namespaces]]` binding, with the create commands documented
- readme: new required setup step, said plainly that this is no longer zero-infra (it was true before, isn't fully true now, said so instead of hiding it)
- test: kv-shaped stub instead of the cache-shaped one, plus a real assertion on the ttl value (mutation-tested it myself — broke the ttl on purpose, confirmed the test catches it, reverted)

also addressed the toctou, not just noted it: kv has no compare-and-swap either, so it doesn't fully go away — documented as an accepted bounded risk in the code comment now, explicit that this is dedup, not exactly-once. and the 5-min window is now commented inline.

`npm test` green, all 6 original cases plus the new ttl check. `wrangler deploy --dry-run` recognizes the kv binding.

PR #941, opened 2026-08-23
feat(harness): add fx (vercel) as a 7th harness

Svector-anu:
## what

fx is vercel's new native (zig) coding agent cli — open-sourced 2026-08-18, [fx.sh](https://fx.sh). `fx ask --json` maps cleanly onto run-harness's existing contract, same shape as the pi/vibe/kimi adapters. genuinely good addition: full native mcp support (pi has none), no npm/pipx install step (native binary), actively maintained by vercel.

everything below is verified against a real, locally-built fx binary (`zig build`, version 0.0.5) — not just read off docs.

## the interesting bits

- **`fx ask --json` prints one json object, not a stream** — simpler than the pi/grok jsonl adapters.
- **usage/cost isn't in that blob.** fx tracks it internally but `ask` doesn't surface it, so the adapter makes a second call, `fx session <id> --json`, keyed off the `session_id` the first call returns. best-effort — a run with real output doesn't fail just because the usage lookup did.
- **no `--model` flag on `ask`.** model selection is `FX_MODEL` (env only, confirmed via `io_mod.getenv`). same for `FX_MAX_AGENT_STEPS`.
- **`--system` on `ask` REPLACES the whole system prompt** (confirmed in `cli_ask.zig` — no append path exists), so compat rules + operator append ride the prompt text itself instead, piped via stdin like `claude.sh`.
- **permission mode is `--auto`, not `--yolo`.** deliberate, not an oversight. grok bypasses its own permission system because grok aborts the whole turn on any denial with zero partial output — a real dealbreaker for unattended use. fx does the opposite: a blocked action fails clean and structured (confirmed live: `{"error":"noninteractive_permission_prompt_unavailable"}`), so there's no reason to skip fx's own real permission system, which is genuinely solid (audited it separately today — allowlist-first admission plus an llm-reviewed second opinion on borderline actions).
- **mcp config shape is NOT the `mcpServers` convention** — fx reads `{home}/.fx/mcp.json` with top-level key `"mcp"` and a different per-server shape. translated in the adapter, tested against the real binary (`fx doctor` picks up the translated config with no complaint).
- **no `FX_HOME` override exists**, so the adapter runs fx under an isolated `HOME` inside `RH_TMPDIR` rather than touching the runner's real one.
- **the one real asymmetry: fx has no openrouter fallback at all** (confirmed — not mentioned anywhere in its own docs). every other harness here treats `OPENROUTER_API_KEY` as the universal default; fx doesn't have that. `install-harness.sh` fails closed on this with a clear named error instead of staging a cli that's guaranteed to fail later inside the actual run. flagged everywhere this matters (resolve-harness.sh, docs, this PR).

## touches, beyond the adapter itself

- `harness-adapter/harnesses.json` — regenerated from the adapter's `rh-meta` block, not hand-edited (matches this repo's own ci gate).
- `scripts/resolve-harness.sh` — allowlist + auth-mode resolution for fx.
- `scripts/install-harness.sh` — stages the curl-installed binary, fails closed on missing credentials.
- `apps/dashboard/lib/harness-auth.ts` — fx's entry in the auth registry. **note:** hand-verified against the interface types (matches the shape of the existing `vibe` entry exactly), but not live type-checked — no `node_modules` in this checkout and a full dashboard `npm install` felt disproportionate for a 13-line addition mirroring a working pattern. worth a real `tsc` pass on your end before merge.
- `docs/harnesses.md` + `harness-adapter/docs/aeon-integration.md` — usage docs, explicitly **not** folded into the existing "all six/all four verified live on 2026-07-22" claims, since fx wasn't part of that sweep and I don't have a real `AI_GATEWAY_API_KEY` to do the equivalent for it.

## verification

- `scripts/tests/test_resolve_harness.sh` — 3 new cases (allowlist reach, both credential vars → native-key). full suite still green.
- `scripts/tests/test_install_harness.sh` — 2 new cases (successful stage, fail-closed on missing creds), with a faked `curl` so it never hits the real network. full suite still green.
- `harness-adapter/adapters/fx.sh` itself, driven directly (no per-adapter test convention exists — `run-harness`'s own dispatch is what `scripts/tests` covers): real missing-credential path against the actual fx binary, real mcp.json translation + `fx doctor` picking it up, real `FX_MODEL`/`FX_MAX_AGENT_STEPS` resolution via `fx doctor`, full success/failure/unparseable-output envelope paths against a stubbed fx recording real argv — and a genuine bug caught and fixed before it shipped: an unbound-variable crash under `set -u` when fx's own output is unparseable (`RESULT`/`EXIT_CODE`/`SID`/`ERR` need default values, not just assignment inside the success path).
- **what's not verified:** an actual live model call through a real `AI_GATEWAY_API_KEY` on github actions. called out explicitly in the docs rather than folded into existing "verified live" language.

## what's honestly missing

a live end-to-end dispatch once someone with a real vercel ai gateway key can run it, and a real `tsc` pass on the dashboard change. flagging both rather than claiming more than i checked.


aaronjmars:
Thanks for this - the `fx.sh` adapter itself is genuinely well-built (correct `emit_envelope` arg order, `set -uo pipefail`, stdin prompt delivery, the RESULT/EXIT_CODE/SID/ERR pre-init to survive `set -u` on the unparseable-output path). `resolve-harness.sh` and `install-harness.sh` are wired correctly and fail closed on missing creds. But it can't land yet - three concrete blockers, all small:

1. **CI is red.** `scripts/tests/test_generate_harnesses_json.sh` hardcodes the six-harness set at line 36 (`[ "$ids" = "claude codex grok kimi pi vibe " ]`) and the count at line 41 (`[ "$c" = "6" ]`). The regenerated `harnesses.json` is now 7, so this parity test fails. Update it to `claude codex fx grok kimi pi vibe` / count 7.

2. **Read-only sandbox is silently skipped on the fx path.** `harness-adapter/run-harness:142` gates the OS sandbox on `case "$HARNESS" in claude | grok | codex | pi | vibe | kimi)` with no `fx` arm and no `*)` default - and run-harness isn't in this PR. The OS sandbox is the sole load-bearing read-only enforcer, so a read-only skill dispatched on fx runs with no sandbox and not even the advisory warning. Add `fx` to that case.

3. **fx can never authenticate on the fleet (DOA end-to-end).** Every other harness credential is enumerated into the job `env:` in `.github/workflows/aeon.yml` (and `messages.yml`), but `AI_GATEWAY_API_KEY` / `VERCEL_OIDC_TOKEN` appear nowhere, including the named-allowlist ALL_SECRETS block (aeon.yml:542). With the secret unset in the job env, `resolve-harness.sh:95` sees it missing, `install-harness.sh` fails closed, and fx never runs. Add the credential to both workflows' env and the ALL_SECRETS allowlist.

Nits (non-blocking): `fx.sh` reads `.exit_code` into `EXIT_CODE` but never checks it (a JSON `exit_code!=0` with empty `error` would still emit a success envelope); MCP translation forces every server to `type:"http"` even though fx also supports `sse`.

Happy to re-review once CI is green and the sandbox + workflow-env wiring is in.

Svector-anu:
all three fixed, no argument on any of them. pushed:

1. `test_generate_harnesses_json.sh` — bumped to `claude codex fx grok kimi pi vibe` / count 7, matches a fresh regen.

2. `run-harness`'s sandbox-gate case now has an `fx` arm. this one had zero test coverage before, which is exactly how it got missed — added `scripts/tests/test_run_harness_sandbox_gate.sh`, which derives the expected harness list straight from `harness-adapter/adapters/*.sh` (not hardcoded, so it can't itself go stale the same way) and asserts every one reaches the sandbox message. mutation-tested it: reverted the `fx` arm, confirmed the test fails with "fx: did NOT reach the sandbox gate" — the exact failure mode you flagged — then reapplied.

3. `AI_GATEWAY_API_KEY`/`VERCEL_OIDC_TOKEN` wired into both workflows. went further than the three sites you named — cross-checked every `MISTRAL_API_KEY` site by hand as a proxy for "any harness-credential env block" and found codex's first implementation pass had missed two more: the "analyze skill output" scorer step in `aeon.yml` (it genuinely needs the credential — re-dispatches through `run-harness` on whatever harness the skill used) and `messages.yml`'s own separate `ALL_SECRETS`-shaped json blob (missed entirely, not partially). both fixed, and I confirmed `MISTRAL_API_KEY` count == `AI_GATEWAY_API_KEY` count in both files now — no more gaps.

nits: `fx.sh` now checks `exit_code` (a json `exit_code!=0` with no `error` field used to still emit a success envelope — now fails with status 3, partial output preserved in the diagnostic). mcp translation preserves an explicit `type:"sse"` instead of forcing every remote server to `http`.

full suite green: `test_generate_harnesses_json.sh`, `test_resolve_harness.sh`, `test_install_harness.sh`, `test_harness_adapter_grok.sh`, `test_harness_envelope.sh`, `test_run_harness_sandbox_gate.sh` (new). both workflow yaml files validated with `yaml.safe_load`.

PR #962, opened 2026-08-25
chore(ci): add eslint + shellcheck lint gates

aaronjmars:
## What

Adds the repo's first lint gates: **eslint** (per app) and **shellcheck** (whole shell surface), plus a `lint` script on every app. Both gates are **green on the current tree**.

## Why

`lint`/`typecheck`/`build`/`test` were never fully normalized and **nothing linted the code** - not the four apps, not the 91 shell scripts that implement the read-only sandbox, harness dispatch, and notify path. This closes that gap without a behavior-change sweep.

## eslint

One flat config per app, folded into the existing per-app `ci-apps.yml` jobs (reuses their installs, keeps the red-X-names-the-surface design):

| app | config | result |
|-----|--------|--------|
| dashboard | `eslint-config-next` 16 **native flat array** | 0 errors / 14 warnings |
| cli | borrows dashboard's eslint + TS rules (same idiom as its typecheck) | clean |
| webhook | `@eslint/js` recommended + Worker globals | clean |
| mcp-server | **skipped** | see below |

- **Next 16 gotcha handled:** wrapping `eslint-config-next` in `FlatCompat` throws `Converting circular structure to JSON` on the react plugin. v16 exports a native flat array, so the config just spreads it. `next lint` was removed in Next 16, so CI runs eslint directly.
- **react-hooks v6:** the new `set-state-in-effect` rule flags 11 pre-existing intentional effects. Kept as `warn` (not `error`) so the gate is green on arrival; ratchet to error after those are refactored.
- **mcp-server skipped on purpose:** `typescript-eslint` hard-refuses its pinned `typescript@7` (upstream typescript-eslint#10940). Not downgrading the build toolchain to satisfy a linter; it joins the gate when typescript-eslint ships TS7 support.

## shellcheck

`scripts/lint-shell.sh` + `ci-shellcheck.yml` gate the whole tracked shell surface (root `aeon`, `scripts/**`, `bin/**`, `harness-adapter/**`, skill helpers) at **error** severity.

- Green after adding a `# shellcheck shell=bash` directive to 7 sourced libs that carry no shebang by design (`capabilities.sh` + 6 `harness-adapter/lib/*`).
- Warning-and-above print as a **non-blocking advisory backlog** in the job log; tighten `GATE_SEVERITY` to `warning` there once it's burned down.
- Portable to macOS bash 3.2 (no `mapfile`); needs no install (shellcheck ships on ubuntu-latest) and no third-party action.

### Latent bugs the advisory backlog surfaced (NOT fixed here - follow-up)

Behavior changes deliberately kept out of this infra PR:

- `scripts/skill-scan.sh:185-188` - `"~/..."` in quotes does **not** expand (SC2088); paths are likely wrong.
- `scripts/audit.sh:39` - overlapping `case` patterns, one never matches (SC2221/SC2222).

## Verification

- `bash scripts/lint-shell.sh` -> exit 0 (91 files); self-clean under shellcheck.
- `npm run lint` in dashboard/cli/webhook -> exit 0.
- `actionlint` clean on both workflows; capabilities-parity + validate-pack + skill_mode tests still pass.

### DIFF
--- a/.github/workflows/ci-apps.yml
+++ b/.github/workflows/ci-apps.yml
@@ -63,6 +63,11 @@ jobs:
         run: npm ci
       - name: Typecheck (strict)
         run: npm run typecheck
+      # eslint-config-next flat config (Next 16 removed `next lint`). Fails only
+      # on error-severity rules; the new react-hooks v6 rules are warnings so the
+      # gate is green on the existing tree (see apps/dashboard/eslint.config.mjs).
+      - name: Lint (eslint)
+        run: npm run lint
       - name: Unit tests
         run: npm test
       # Kept separate from the typecheck on purpose — see the TS7 note above.
@@ -94,6 +99,11 @@ jobs:
         working-directory: apps/cli
       - name: Typecheck (strict, includes the shared dashboard lib)
         run: npm run typecheck
+        working-directory: apps/cli
+      # Borrows the dashboard's eslint + TypeScript rules (installed above), the
+      # same single-source-of-truth idiom as the typecheck.
+      - name: Lint (eslint, borrows the dashboard install)
+        run: npm run lint
         working-directory: apps/cli
 
   mcp-server:
@@ -135,6 +145,10 @@ jobs:
       # No lockfile committed for this app either — see mcp-server above.
       - name: Install deps
         run: npm install --no-audit --no-fund
+      # Plain-JS Worker: @eslint/js recommended with Worker runtime globals
+      # declared (see apps/webhook/eslint.config.mjs).
+      - name: Lint (eslint)
+        run: npm run lint
       # --dry-run validates wrangler.toml and bundles the Worker without any
       # Cloudflare credentials, so a broken config or a bad import fails here
       # instead of at deploy time.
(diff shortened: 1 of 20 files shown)


aaronjmars:
### Added: root task aggregator (F2)

`package.json` at the repo root gives one command from anywhere:

```
npm run bootstrap   # install every app (in the CLI-borrow order)
npm run lint        # dashboard + cli + webhook eslint
npm run lint:shell  # scripts/lint-shell.sh
npm run typecheck   # dashboard + cli + mcp-server
npm run build       # dashboard + mcp-server
npm run test        # dashboard + webhook
```

**Deliberately NOT an npm workspace** - no `workspaces` key, so nothing hoists or dedupes. A real workspace would break three things this repo relies on: the divergent `typescript` pins (dashboard@6 vs mcp-server@7), the four standalone per-app lockfiles Dependabot updates per directory, and the CLI's `../dashboard/node_modules/.bin` borrow. The scripts just fan `npm --prefix` across the apps.

Verified the parent `package.json` does not trigger Next's workspace-root inference (dashboard build still green); root `lint` / `typecheck` / `lint:shell` all exit 0. It touches no path filter, so it triggers no new CI on its own - the apps' existing jobs already validate them.

PR #968, opened 2026-08-26
add machine-readable vuln scanner execution evidence

Svector-anu:
## why

A vuln-scanner run can stage Semgrep, TruffleHog, OSV-Scanner, Slither, and cargo-fuzz successfully while the harness report says they were skipped. Model prose cannot prove whether a scanner actually ran.

## fix

- wrap staged scanner binaries with a tiny execution logger
- add a post-run workflow step that prints the staged manifest and actual invocation log
- warn explicitly when no scanner invocation evidence exists

This keeps the existing best-effort behavior and does not turn an absent optional scanner into a workflow failure.

## verification

- `git diff --check` passed
- shellcheck is not installed in the local environment
- branch diff is limited to `scripts/stage-vuln-scanner.sh` and `.github/workflows/aeon.yml`
- real Cursor run `32931159345` demonstrated the evidence gap: staging succeeded, but the agent report claimed all scanners were skipped

### DIFF
--- a/.github/workflows/aeon.yml
+++ b/.github/workflows/aeon.yml
@@ -1051,6 +1051,17 @@ jobs:
 
       # Egress hardening P2: publish the warn-mode audit log. always() so it lands
       # even if the run failed; the host list is the allowlist oracle P3 builds from.
+      - name: Upload egress audit log
+        if: always() && vars.EGRESS_AUDIT == '1'
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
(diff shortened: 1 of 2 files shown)


Svector-anu:
Runtime follow-up (2026-08-26): the execution-evidence instrumentation was exercised in real Actions runs.

- no-eligible-target run `32935236179`: staged tools installed; execution log recorded `executions=none` rather than implying scans ran.
- explicit Hermes scan rerun `32940739251`: per-tool evidence reported `semgrep=not-observed`, `trufflehog=not-observed`, `osv-scanner=not-observed`, and `slither/cargo-fuzz=not-observed-or-not-applicable`.

This is why the provisional dependency claims remain unpromoted: the report is not treated as scanner-verified without an invocation record.

PR #967, opened 2026-08-25
add cursor hermes and glm harnesses

Svector-anu:
## summary

Add three adapter-backed harnesses to Aeon:

- Cursor CLI (`agent -p`) with `CURSOR_API_KEY`
- Hermes Agent via Nous Portal (`hermes -z`, `HERMES_AUTH` OAuth archive)
- GLM Coding Plan via Claude Code against Z.AI's documented Anthropic endpoint (`GLM_API_KEY` / `ZAI_API_KEY`)

The implementation follows the existing fx/codex/pi/vibe/kimi `run-harness` contract and wires resolver, installer, both workflows, dashboard harness/model/auth registries, MCP server dispatch, capability manifest, and regression tests.

## verification

- `bash scripts/tests/test_resolve_harness.sh`
- `bash scripts/tests/test_install_harness.sh`
- `bash scripts/tests/test_generate_harnesses_json.sh`
- `bash scripts/tests/test_run_harness_sandbox_gate.sh`
- `bash scripts/tests/test_harness_envelope.sh`
- `npm --prefix apps/dashboard run typecheck`
- `npm --prefix apps/dashboard test` (195 passing)
- `npm --prefix apps/mcp-server run typecheck`
- `git diff --check`

Provider documentation used:

- https://cursor.com/docs/cli/headless
- https://hermes-agent.nousresearch.com/docs/reference/cli-commands
- https://hermes-agent.nousresearch.com/docs/integrations/nous-portal
- https://docs.z.ai/devpack/quick-start
- https://docs.z.ai/devpack/tool/others

Real provider-backed Actions runs require the corresponding repository secrets (`CURSOR_API_KEY`, `HERMES_AUTH`, and `GLM_API_KEY`/`ZAI_API_KEY`); the current fork has none of those three yet, so no successful live provider run is claimed here.

PR #975, opened 2026-08-27
fix dashboard auth rows for new harnesses

Svector-anu:
## What

Make the Cursor, Hermes, and GLM credentials permanent first-class rows in Settings → Access Keys, and give Hermes the same Connect/Reconnect behavior as the existing captured-login harnesses.

## Why

PR #967 added all three harnesses to installation, resolution, execution, and the top-bar picker, but their credentials were not added to the dashboard builtin catalog. A secret appeared only after it had already been set, as a generic custom Skill Keys row; an unset credential disappeared entirely. HERMES_AUTH also was not mapped to its native OAuth flow, so its set row offered Remove but no Reconnect.

## Fix

- add CURSOR_API_KEY, HERMES_AUTH, and GLM_API_KEY as permanent Core credentials with setup context
- derive OAuth-secret → harness mappings from the central harness registry, covering Hermes and preventing future parallel-list drift
- add Cursor, Nous, and Z.AI service marks
- add accurate modal titles/copy and only mention OpenRouter fallback for harnesses that actually support it
- add focused regression coverage for permanent visibility and captured-login routing

## Verification

- dashboard tests: 198 passed
- dashboard typecheck: passed
- dashboard production build: passed (one existing Turbopack trace warning)
- dashboard lint: 0 errors, 14 existing warnings
- live metadata smoke against Svector-anu/svectors-lab: all three returned group=Core and isSet=true; no secret values were read
- git diff --check: clean

Mutation checks:

1. Removed only the three builtin catalog rows while keeping the test: `keeps every new harness credential visible even before it is set` failed with actual `undefined`, expected `Core`.
2. Excluded only Hermes from the derived OAuth mapping while keeping the test: `maps every captured login secret to its Connect/Reconnect flow` failed because HERMES_AUTH was absent.
3. Restored both fixes: focused 12/12 green and full dashboard suite 198/198 green.

### DIFF
--- a/apps/dashboard/components/HarnessAuthModal.tsx
+++ b/apps/dashboard/components/HarnessAuthModal.tsx
@@ -4,16 +4,25 @@ import { useState } from 'react'
 import { inputCls } from '../lib/utils'
 import { HARNESS_AUTH } from '../lib/harness-auth'
 
-// Native auth for the run-harness harnesses (codex/pi/vibe/kimi) — the generic
+// Native auth for the adapter-backed harnesses — the generic
 // parallel to GrokAuthModal. codex/kimi offer a one-click native login (browser/
-// device OAuth, captured for CI); every one of the four also takes a provider
-// API key, and all fall back to the shared OpenRouter key. Posts to
+// device OAuth, captured for CI); others take a provider API key. Some harnesses
+// also fall back to the shared OpenRouter key. Posts to
 // /api/harness-auth via onHarnessAuth (no arg = OAuth, {key} = provider key).
 
-const TITLES: Record<string, string> = { codex: 'Codex', kimi: 'Kimi', pi: 'Pi', vibe: 'Vibe' }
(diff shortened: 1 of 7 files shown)


aaronjmars:
Merging this - correct, tested, and green. One follow-up nit (not a blocker, predates this PR): the `fx` harness (added in #941; `AI_GATEWAY_API_KEY`) has the exact same catalog gap this PR fixes for cursor/hermes/glm - it only shows as a generic Custom-secret row once set, and is invisible when unset. Worth a one-line `AI_GATEWAY_API_KEY` Core row in `secrets-catalog.ts`, plus a registry-derived parity test (iterate `HARNESS_AUTH` native-key secrets against `BUILTIN_NAMES`) so the next harness can't silently drift out again. Also the `OPENROUTER_API_KEY` description in secrets-catalog.ts:19 now undercounts - hermes falls back to OpenRouter too, so it's five harnesses, not four. Follow-up PR is fine.

PR #986, opened 2026-08-30
fix(dashboard): require exact origin host match

keyurbodar:
## Summary

Require state-changing dashboard requests to use the same host in the Origin/Referer header and the request Host header.

## Problem

The API gate allowed loopback aliases independently. A browser page from `http://localhost:5555` could send a no-CORS POST to `http://127.0.0.1:5555`, and both hosts passed the allowlist even though they are different browser origins.

This exposed privileged routes such as `/api/secrets`, `/api/auth`, and `/api/skills/<name>/run` to cross-origin requests between loopback aliases.

## Fix

- Normalize and validate the request Host header.
- Require the Origin/Referer host to match it exactly, including port.
- Preserve safe-method handling, configured allowed hosts, and `AEON_DASHBOARD_ALLOW_ANY_HOST=1`.
- Add regression coverage for mismatched loopback aliases, exact host matches, malformed hosts, and Referer fallback.

## Testing

- `npm test -- --test-name-pattern='(isSameOriginWrite|gateRequest)'`
- Result: 202 tests passed

### DIFF
--- a/apps/dashboard/lib/security/api-gate.test.ts
+++ b/apps/dashboard/lib/security/api-gate.test.ts
@@ -88,36 +88,83 @@ describe("isSameOriginWrite", () => {
   });
   it("POST with same-origin Origin passes", () => {
     assert.equal(
-      isSameOriginWrite("POST", headers({ origin: "http://localhost:5555" })),
+      isSameOriginWrite(
+        "POST",
+        headers({ host: "localhost:5555", origin: "http://localhost:5555" }),
+      ),
       true,
     );
     assert.equal(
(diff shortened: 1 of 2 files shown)


aaronjmars:
Reviewed - solid CSRF/alias hardening, merging.

One nit for the record (non-blocker, no change needed): theoretical https default-port asymmetry. A `Host` of `localhost:443` normalizes to `localhost:443` (the http reparse keeps 443), while an `Origin` of `https://localhost:443` strips to `localhost` (443 is the https default), so the exact-match compare would reject a same-origin request. Irrelevant for an http loopback dashboard, so skipping - noting it only in case the gate is ever reused behind TLS.

PR #1001, opened 2026-08-31
docs: optional runtime hardening for the Aeon MCP server

alexverify:
## What

Adds a "Runtime hardening (optional)" section to `docs/skill-integrity.md`. It documents eyebrow wrap as an operator side complement to the ci-skill-integrity gate, for the Aeon MCP server that `bin/add-mcp` registers with Claude Code.

## Why

The CI gate protects the skills in the repo at PR time. It says nothing about the server once it runs on an operator machine. The new section explains the opt in local relay (per tool policy, redacted audit log, tool surface capture, OS sandbox) that covers that surface. Docs only. No change to install or CI paths, and `eyebrowlock.json` is untouched.

 FYI, separate from this PR: eyebrow v0.4.2 is out and your gate pins v0.4.1. Bump when convenient. I left it as is here to keep this change focused.

## Type of change
- [x] Core fix (dashboard / scripts / workflows / docs)

---
### Core fix (dashboard / scripts / workflows / docs)

- [x] Change is focused — one concern, no unrelated refactor
- [x] Touched code follows the existing pattern in the file
- [x] Relevant CI gates pass locally (e.g. `bash scripts/check-skill-categories.sh`, `bash scripts/check-capabilities-parity.sh`)
- [x] Touched `apps/**`? That app typechecks, tests, and builds (dashboard: `npm run typecheck && npm test && npm run build`)

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,12 @@ from or pin to; the template keeps serving the latest `main` to new forks.
 
 ### Added
 
+- **Docs: optional runtime hardening for the installed MCP server.**
+  [`docs/skill-integrity.md`](docs/skill-integrity.md) now describes `eyebrow
+  wrap`, an operator-side complement to the CI gate that routes the registered
+  Aeon MCP server through a local relay for per-tool policy, a redacted audit
+  log, tool-surface capture, and OS-sandbox confinement. Opt-in; no change to the
+  install or CI paths.
 - **Telegram notifications reply to the previous run of the same skill.** Default on. Ledger at `memory/telegram-threads/<skill>.json`. Kill switch: repo variable `TELEGRAM_REPLY_TO_PREVIOUS=0`. (#995)
 - **Three more run-harnesses: Cursor, Hermes, and GLM.** Cursor CLI (`agent -p`,
   `CURSOR_API_KEY`), Hermes via the Nous Portal (`hermes -z`, `HERMES_AUTH`), and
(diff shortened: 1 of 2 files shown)


aaronjmars:
Hold. The gap is real (CI fingerprints skills in the repo, not the live local MCP server), but this section as written will send operators down the wrong path.

**Wrong config surface.** `bin/add-mcp` is `claude mcp add` with default local scope, which writes `~/.claude.json` under the project path. This repo has no root `.mcp.json`. Wrap will still see a default `add-mcp` install because it also reads that local store; the sentence that says "for an Aeon server declared in a project `.mcp.json`" points at a file this install path does not use. User-scope needs `eyebrow wrap --global`.

**OS sandbox is sold as a win for this server.** `apps/mcp-server` spawns a harness (`claude` by default) as a child. Default wrap enables Seatbelt/bwrap with writes limited to workspace+temp and, on macOS, network limited to the local proxy port. That child wants `~/.claude` writes and API egress. eyebrow's own usage doc already warns some servers need `--workspace` / `--no-sandbox`. Following this page as written can break `aeon-*` skill runs.

Also drop "the same engine" (scan/verify vs wrap/shim are different components), stop claiming wrap detects a binary swapped outside the build, and link the actual wrap runbook (`docs/usage.md`) with `eyebrow wrap` / `eyebrow wrap --status` / `eyebrow unwrap`.

Docs-only, no CI to fail. Please rewrite the section to match how Aeon actually installs MCP, with the sandbox caveat, then this is mergeable.

PR #1029, opened 2026-09-05
fix(dry-run): support explicit claude and codex model selection

Svector-anu:
## what

let local skill dry runs select claude or codex and a native model id through the existing `harness-adapter/run-harness` contract. no new harness, model pin, account configuration or dashboard change. claude remains the default.

## why

while building a small procedural three.js city game (walking, driving, delivery and toy-brick characters) in a supervised codex session, we tried to validate its associated aeon skill locally. that exposed a portability gap: `scripts/dry-run.sh` unconditionally launches `claude -p`, even when the operator wants to test through codex with `gpt-6-astra`.

the game itself is not evidence of an autonomous aeon build. its earlier full skill attempt timed out at 180 seconds; we have not established that timeout's underlying cause. the demonstrated bug is the lack of harness/model selection, independently visible in current upstream source.

parity checked against upstream `048dee9111e311cf9dc54e153876fcbd86e98244`: both pre-fix dry-run source and its existing test are byte-identical to the fork baseline. upstream's last change to the source is `a95f09e7` (#914). an operator selecting codex for a local skill check still gets claude instead, so the gate cannot validate their chosen harness.

## fix

- add `--harness` / `--model`, with `DRYRUN_HARNESS` / `DRYRUN_MODEL` defaults; reject unsupported harnesses and known wrong-family model ids rather than silently switching.
- use the canonical `skill_mode.sh mode <skill>` and `allowed-tools <mode>` interfaces. the old `<skill>` call fell back to `Read,Bash`; nested `metadata.mode` was also missed.
- preserve synthetic task credentials and notification capture; require the read-only os wrapper and a timeout utility.
- reject missing cli, invalid/error envelopes, empty results and raw-output wrapper fallbacks. stderr alone is not useful output.
- record harness, requested model, exit, duration, result length and reason. invalidate stale receipts before invocation.
- handle empty requirements on macos bash 3.2 and encode them as a real empty json list.

the existing gate is not a complete credential sandbox: local profiles, ssh/helpers and undeclared environment credentials remain outside the synthetic-token guarantee. this limitation is documented, not hidden. no safety bypass flags were added.

## verification

on the clean upstream-based branch:

```text
bash scripts/tests/test_dry_run.sh
dry-run: 19 passed, 0 failed
dry-run dispatch: 35 passed
```

the existing `ci-tests.yml` step already invokes `test_dry_run.sh`; it now calls the new offline dispatch suite. bash syntax and `git diff --check` are clean. shellcheck is unavailable locally. github ci results will be linked separately after this opens; not claimed in advance.

three source mutations were rerun on this upstream-based branch, one at a time:

```text
remove model forwarding:
FAIL: grep -qx -- --model output/args
restore the old tool-helper invocation:
FAIL: grep -q Bash(./notify: output/args
remove raw-wrapper rejection:
FAIL: expected rejection (reply=wrapped, args=--harness codex)
```

restored all three source changes and reran the green suite.

real local model probes on the fix, using the actual shared dispatcher and adapters, existing cli logins, an isolated read-only fixture, no requested tools and a 90-second ceiling:

```text
dry-run: harness=codex model=gpt-6-astra mode=read-only timeout=90s
dry-run: harness=codex exit=0 elapsed=11s result_chars=31 reason=ok
dry-run: harness=claude model=sonnet mode=read-only timeout=90s
dry-run: harness=claude exit=0 elapsed=41s result_chars=31 reason=ok
```

both returned exactly `AEON_DRYRUN_ROUTING_OK_20260905`. those are live routing/response receipts, not a comparative speed benchmark or independent server-side model attestation. the first nested-sandbox attempt was rejected with exit 71; approved execution outside the parent sandbox passed with aeon's own read-only wrapper still enabled. the empty-requires failures discovered during those attempts are covered by regression tests.

the separate game currently passes 8/8 unit tests and builds (rerun locally). its saved browser receipt records a completed delivery and zero page errors, but the updated characters measured 28.4 fps in that run, below the 60 fps target. no game source, screenshots, private repo data or self-healing claim is included in this framework patch.

## scope

fresh branch from upstream/main, one isolated commit, four intended files. no `memory/`, disclosure material, `aeon.yml`, credentials, `.mcp.json`, game assets or unrelated fork history.

- [x] core fix; one local dry-run invocation concern
- [x] shared dispatcher and canonical mode helpers reused
- [x] focused tests and mutation checks pass locally
- [x] no `apps/**` changes

### DIFF
--- a/scripts/dry-run.sh
+++ b/scripts/dry-run.sh
@@ -25,11 +25,11 @@
 #       Fail if any provided real secret value appears in env-file. The guarantee
 #       the whole feature rests on.
 #
-#   run <skill> [var]
+#   run <skill> [var] [--harness claude|codex] [--model <native-id>]
 #       Orchestrate a full dry run: build a synthetic env for the skill's requires,
 #       route ./notify to a capture file (NOTIFY_DRY_RUN), stub git/gh with a fake
-#       token so any push/write fails auth and is discarded, run the skill through
-#       the claude harness under a tight time/turn ceiling, then evaluate. Emits the
+#       token (not a complete remote-write sandbox), run the skill through
+#       shared harness dispatcher under a wall-clock ceiling, then evaluate. Emits the
 #       verdict to $DRYRUN_VERDICT (default output/.dry-run/<skill>.json).
 #
 # Gate toggle: SKILL_DRYRUN repo variable, default on. A fork that hits a false
@@ -42,7 +42,7 @@ ROOT="$(cd "$HERE/.." && pwd)"
 # Keys that must NEVER be synthesized: the model auth the dry run genuinely needs.
 is_model_key() {
   case "$1" in
-    ANTHROPIC_API_KEY|ANTHROPIC_OAUTH_TOKEN|CLAUDE_CODE_OAUTH_TOKEN) return 0 ;;
+    ANTHROPIC_API_KEY|ANTHROPIC_OAUTH_TOKEN|CLAUDE_CODE_OAUTH_TOKEN|OPENAI_API_KEY|CODEX_AUTH) return 0 ;;
     *) return 1 ;;
   esac
 }
@@ -68,6 +68,7 @@ synth_value() {
 
 cmd_synth_env() {
   local csv="${1:-}"
+  [ -n "$csv" ] || return 0  # bash 3.2 treats an empty array as unset under -u
   IFS=', ' read -ra keys <<< "$csv"
   for key in "${keys[@]}"; do
     key="${key%\?}"                     # strip the `?` works-better marker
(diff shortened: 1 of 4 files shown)


Svector-anu:
verified the actual github runs for head `14980edac3a772f577c94acb837d040ded7664f0`:

- [ci-tests passed](https://github.com/aeonfun/aeon/actions/runs/33954503810), including the registered dry-run step. literal lines from its log:

  ```text
  dry-run: 19 passed, 0 failed
  dry-run dispatch: 35 passed
  ```

- [shellcheck error-severity gate passed](https://github.com/aeonfun/aeon/actions/runs/33954503801).

the `REAL secret value leaked` diagnostic in the test log is expected from the negative test using the literal synthetic fixture `real-token-abc123-do-not-leak`; it does not indicate a live credential was used or leaked.

also rechecked github's actual pr files: one commit, four files, +294/-28. no memory, personal config, game files or unrelated fork changes. this is an open proposal, not a merge.

scripts/dry-run.sh
```diff
@@ -42,7 +42,7 @@ ROOT="$(cd "$HERE/.." && pwd)"
 # Keys that must NEVER be synthesized: the model auth the dry run genuinely needs.
 is_model_key() {
   case "$1" in
-    ANTHROPIC_API_KEY|ANTHROPIC_OAUTH_TOKEN|CLAUDE_CODE_OAUTH_TOKEN) return 0 ;;
+    ANTHROPIC_API_KEY|ANTHROPIC_OAUTH_TOKEN|CLAUDE_CODE_OAUTH_TOKEN|OPENAI_API_KEY|CODEX_AUTH) return 0 ;;
```
Svector-anu, line 45: [CRITICAL] Exempt model credentials by the selected harness instead of treating both families as globally safe. With this list, a Claude dry run inherits the caller’s real OPENAI_API_KEY/CODEX_AUTH (and a Codex run inherits real Anthropic credentials), so candidate skill code can read or exfiltrate an unrelated live credential instead of receiving a synthetic value — why it matters: the security gate can expose production model credentials that the chosen harness does not need.

scripts/dry-run.sh
```diff
@@ -141,23 +142,60 @@ cmd_assert_no_real_secrets() {
...
+  [[ "$skill" =~ ^[a-z0-9][a-z0-9-]*$ ]] || { echo 'dry-run: invalid skill name' >&2; return 2; }
+  cd "$ROOT"
+  local verdict="${DRYRUN_VERDICT:-output/.dry-run/${skill}.json}"
+  mkdir -p "$(dirname "$verdict")"
+  reject() { jq -cn --arg reason "$1" '{passed:false,reasons:[$reason]}' | tee "$verdict"; return 1; }
+  # Invalidate any previous receipt before parsing or invoking dependencies.
+  jq -cn '{passed:false,reasons:["dry run not completed"]}' > "$verdict"
+  # Explicit flags win over dry-run env; no implicit model/family fallback.
+  local harness="${DRYRUN_HARNESS:-claude}" model="${DRYRUN_MODEL:-}"
+  if [ $# -gt 0 ] && [[ "$1" != --* ]]; then var="$1"; shift; fi
```
Svector-anu, line 155: [ISSUE] Preserve a positional var that starts with `--` (or add an explicit separator/input option) before parsing harness flags. Inputs such as `dry-run.sh run pr-review --survey` are now interpreted as unknown options — why it matters: valid skills whose documented input grammar begins with a flag can no longer be exercised by this gate.

Svector-anu (commented):
**Verdict**: blocked: unrelated live model credentials can reach candidate skill code
The dry-run must scope real model auth to the selected harness, and its new option parser must retain flag-shaped skill inputs.

- [CRITICAL] scripts/dry-run.sh:45 — globally exempting both credential families exposes unrelated real model credentials to candidate skill code instead of synthesizing them.
- [ISSUE] scripts/dry-run.sh:155 — treating every leading `--` token as a runner option prevents valid flag-shaped skill inputs such as `pr-review --survey` from being dry-run.

PR #1039, opened 2026-09-08
feat(vuln-scanner): add Riva research kernel

Svector-anu:
## Summary
Adds Riva as an opt-in focused vulnerability-research kernel for Aeon’s vuln-scanner while preserving legacy behavior.

## Changes
- compact target dossier and focused threat-model/invariant research kernel
- legacy, shadow, and Riva selectors
- fail-closed candidate validation and regression tests
- shadow mode is comparison-only, read-only, disables MCP, and withholds disclosure credentials

## Compatibility
The existing scanner, PoC, disclosure, tracking, memory, and safety features remain unchanged. Riva is opt-in until shadow evidence is reviewed.

## Validation
Local context, kernel-contract, candidate-schema, shadow-guard, comparison, workflow, and existing vulnerability-gate tests pass.


aaronjmars:
Thanks @Svector-anu - solid, security-conscious addition. Riva/shadow modes are fail-closed and read-only, the legacy scanner default is untouched, and the skills.json + packs.json regen is correct (parity checks green).

One real blocker before merge:

**`verify` / `ci-skill-integrity` is red** - `eyebrow verify --ci` fails with drift because the PR edits `skills/vuln-scanner/SKILL.md` and adds `riva.md` but does not update `eyebrowlock.json`. Please run:

```
eyebrow scan --path . --lockfile eyebrowlock.json
```

and commit the result. That is the only failing check; everything else (shellcheck, tests, catalog parity) is green.

Two smaller things worth a look while you are in there:

1. Shadow/compare mode unsets `GH_TOKEN`/`GH_GLOBAL` (aeon.yml:917), but the scan arm still uses `gh repo fork --clone` and `gh api search/repositories` to reach the target. A shadow run may fail to fork/clone with no GitHub creds (a plain public `git clone` works, the `gh` paths will not). Since aeon.yml is not exercised by PR CI, this is only hit on the first live run - worth confirming shadow can actually reach a target before relying on it.
2. `scripts/resolve-riva-capabilities.sh:4` sets `set -u`, which leaks into the sourcing run block. Low risk given the `${VAR:-}` usage, but a sourced helper should not mutate the caller shell options. Also `RIVA_SHADOW_MODE` written to `$GITHUB_ENV` at :16 is never read.

Flip to a merge once eyebrowlock is regenerated. Nice work.

scripts/resolve-riva-capabilities.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+# Source from the Aeon runner after SKILL_NAME and SKILL_VAR are exported.
+# Resolves the capability tier before MCP or skill-secret setup.
+set -u
+
+SHADOW_MODE=0
+if [ "${SKILL_NAME:-}" = "vuln-scanner" ] && [[ "${SKILL_VAR:-}" == shadow:* || "${SKILL_VAR:-}" == compare:* ]]; then
```
Svector-anu: [CRITICAL] Match the documented bare `shadow` and `compare` selectors here as well (for example, test exact values in addition to `shadow:*` / `compare:*`). `skills/vuln-scanner/SKILL.md` accepts bare `shadow` and `compare`, but this guard leaves those runs in write mode and the later workflow injects their declared credentials, so a mode advertised as comparison-only can retain mutation and disclosure capability.

Svector-anu (commented):
**Verdict**: blocked: bare shadow selectors bypass isolation
The comparison-only guarantee does not hold for two documented invocation forms.

- [CRITICAL] scripts/resolve-riva-capabilities.sh:7 — Bare `shadow` / `compare` selectors remain write-capable and receive declared credentials, so comparison-only runs can retain mutation and disclosure capability.

aaronjmars:
Confirmed your `blocked` verdict - agreed, this is a real fail-open, not a nit.

The guard in `scripts/resolve-riva-capabilities.sh:7` only trips `SHADOW_MODE=1` when `SKILL_VAR` matches `shadow:*` / `compare:*` (colon required), but `SKILL.md:417` routes the BARE `shadow` / `compare` selectors (no colon) to the same kernel. So the documented bare invocations run shadow logic while keeping full write tools + injected creds (GH_TOKEN, RESEND_*, XAI_API_KEY) - the "read-only, comparison-only, credentials withheld" guarantee is void for exactly the documented forms.

`scripts/tests/test_riva_shadow_guard.sh` is grep-only (asserts strings exist in files), so CI stays green over the hole - false assurance.

To unblock:
1. Fix the `:7` guard to also catch the bare selectors, e.g. match `shadow|shadow:*|compare|compare:*` (or key off the resolved `KERNEL` value) so it aligns with the `SKILL.md` case block.
2. Make `test_riva_shadow_guard.sh` behavioral: invoke the resolver with a bare `shadow` selector and assert `SHADOW_MODE=1` + creds stripped, instead of grepping.

Re-request review once that lands. Skill registration (skills.json / packs.json / eyebrowlock) looks complete - thanks for the clean self-review.

Svector-anu:
Addressed the review blockers and rebased via merge onto current upstream main:

- bare `shadow` and `compare` now enter the same fail-closed isolation path as `shadow:*` / `compare:*`
- resolver strips ambient GitHub, disclosure, and model credentials immediately at the capability boundary
- removed sourced `set -u` side effect and the unused `RIVA_SHADOW_MODE` environment write
- replaced the grep-only guard test with behavioral coverage for all four selector forms, credential stripping, read-only mode, caller shell options, and normal-scan non-regression
- shadow target cloning now uses public `git clone`, so it does not depend on withheld GitHub credentials
- bare shadow without chained target data exits cleanly and asks for `shadow:owner/repo`, avoiding unauthenticated `gh` discovery
- regenerated catalogs and the Riva integrity baseline

All 7 upstream checks are green and the PR is clean/mergeable. Re-review requested.

Svector-anu (commented):

aaronjmars:
Thanks for the thorough rework. I verified all seven fixes against the diff and they are genuinely done and correct:

- bare `shadow`/`compare` now route through the same fail-closed isolation path as the `:*` forms (resolver and SKILL.md agree)
- credentials stripped at the capability boundary (ALL_SECRETS/GH_*/RESEND_*/XAI_API_KEY)
- the sourced `set -u` side effect and the dead `RIVA_SHADOW_MODE` write are removed
- the guard test is now behavioral across all four selector forms plus a non-regression case
- shadow cloning uses public `git clone`
- bare `shadow` with no target exits cleanly asking for `shadow:owner/repo`
- catalogs and eyebrowlock regenerated, all checks green

Code, tests, and skill registration are merge-ready. One blocker before I can land it: the PR also commits a live scan run's artifacts into this public repo, and some of that must not be public.

Please `git rm` these 9 files and force-push (keep only the skill/code/test/catalog/eyebrowlock changes):

- `output/articles/vuln-scan-2026-09-10.md` and `output/articles/vuln-scan-2026-09-10.notify.md` - these write up a provisional HIGH candidate in a real third-party project with an exact PoC path. The article itself says it should stay private until verified and disclosed; committing it here publishes it before verification and before the project is contacted, which is the opposite of our disclosure policy.
- `memory/telegram-threads/vuln-scanner.json` - contains a real Telegram chat_id and message_id.
- `memory/logs/2026-09-10.md`, `memory/vuln-scanned.json`, `memory/skill-health/vuln-scanner.json`, `memory/token-usage.csv`, `apps/dashboard/outputs/vuln-scanner-2026-09-10T14-53-13Z.json`, `output/.chains/vuln-scanner.md` - per-run state that does not belong in canon.

Once those are out, this is good to merge. Thanks again.

Svector-anu:
Done. Dropped the `chore(vuln-scanner): auto-commit 2026-09-10` commit entirely (it was exactly those 9 run-artifact files and nothing else) and force-pushed — branch is now at the `riva-upstream-clean` merge with current upstream main, carrying only the skill/code/test/catalog/eyebrowlock changes and the 7 verified fixes.

Confirmed via `gh pr view --json files` that none of the flagged paths remain in the diff:
- output/articles/vuln-scan-2026-09-10.md + .notify.md
- memory/telegram-threads/vuln-scanner.json
- memory/logs/2026-09-10.md, memory/vuln-scanned.json, memory/skill-health/vuln-scanner.json, memory/token-usage.csv
- apps/dashboard/outputs/vuln-scanner-2026-09-10T14-53-13Z.json
- output/.chains/vuln-scanner.md

Head is now e249ecd, still MERGEABLE.

PR #1053, opened 2026-09-10
fix(chain): distinguish no-action telemetry

Svector-anu:
## bug

A completed dev-loop with no verified feature PR currently emits `CHAIN_STATUS=success`. That makes a no-op indistinguishable from a chain that shipped and passed review, so skill-health can overstate dev-loop reliability.

## fix

- emit `CHAIN_STATUS=no-action` when `CHAIN_NO_ACTION=true`
- exclude `no-action` from cron-state reliability bookkeeping, alongside `invalid-dispatch`
- leave genuine `failed` and reviewed `success` paths unchanged
- add an anchored-extraction regression test and wire it next to the invalid-dispatch test

The Actions run remains visible as a completed run with a distinct status, but it counts as neither a functional success nor a functional failure.

## proof

- post-merge fork proof on `Svector-anu/svectors-lab@4617511a`: both no-action and invalid-dispatch regression tests pass; workflow YAML parses
- upstream port test: pass
- mutation: restoring the old success assignment fails with `no-action emitted success` and `no-action reached cron-state write`
- restored fix: both tests pass and YAML parses
- `git diff --check`: pass
- local lint wrapper could not invoke shellcheck because shellcheck is not installed; upstream CI provides the shellcheck gate

Port of Svector-anu/svectors-lab#75, following the fork-to-upstream pattern used by #1051/#1052.

### DIFF
--- a/.github/workflows/chain-runner.yml
+++ b/.github/workflows/chain-runner.yml
@@ -432,7 +432,7 @@ jobs:
             echo "CHAIN_STATUS=failed" >> "$GITHUB_ENV"
           elif [ "$CHAIN_NO_ACTION" = "true" ]; then
             echo "Chain '$CHAIN' completed without a review: feature produced no verified PR"
-            echo "CHAIN_STATUS=success" >> "$GITHUB_ENV"
+            echo "CHAIN_STATUS=no-action" >> "$GITHUB_ENV"
           else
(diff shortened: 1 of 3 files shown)


Svector-anu (commented):

PR #1054, opened 2026-09-10
fix(vuln-scanner): bound the trufflehog filesystem scan, forbid backgrounding scanners

Svector-anu:
## Problem

The `trufflehog git` (history) pass is bounded with `timeout 300`. The `trufflehog filesystem` pass is **not**.

On a very large working tree — a big monorepo checkout — the filesystem walk burns the whole turn budget by itself. Same failure mode the history walk had before it was bounded:

- the run reports **`success`** with no report written and no ledger entry
- two `.trufflehog.real` processes are still alive at job cleanup (`Terminate orphan process`)
- the agent's final output is a placeholder: *"Waiting for the background scan tasks to complete — I'll pick back up automatically when they finish"* — which a one-shot `workflow_dispatch` run can never honor

Observed live 2026-09-10 on a real large-repo scan (Azure CLI).

## Fix

Minimal, mirrors the existing git-history handling:

- wrap `trufflehog filesystem` in `timeout 300`, keep the existing `|| TRUFFLEHOG_RC=$?` capture, emit the same `VULN_SCANNER_TIMEOUT` notice
- `sources.txt` records `trufflehog=timeout` on RC 124 (matching `trufflehog-git=timeout`)
- propagate `trufflehog=<ok|fail|timeout>` into the A7 report line, the clean/limited-audit line, and the `Mode: scan` log template
- extend the durable-line note so it covers both timeout fields
- comment in the block: run the scanner commands verbatim; a hand-written invocation still needs the `timeout` wrapper and **must not be backgrounded with `&`** — a scan still running when the report is written is a scan that didn't finish

## Verification

- `bash -n` clean on the A3 scanner block
- diff is one file, ~20 lines, no behavior change to the happy path (a scan that completes under 300s is byte-identical in output)

### DIFF
--- a/skills/vuln-scanner/SKILL.md
+++ b/skills/vuln-scanner/SKILL.md
@@ -162,9 +162,14 @@ fi
 
 # --- Secrets: TruffleHog (only-verified = actually authenticates) ---
 if command -v trufflehog >/dev/null 2>&1; then
+  # BOTH passes are bounded. Run these verbatim; if you invoke trufflehog (or any
+  # scanner) yourself, it still needs the `timeout 300` wrapper and must not be
+  # backgrounded with `&` — a scan still running when you write the report is a
+  # scan you did not finish, and a one-shot workflow_dispatch run cannot resume it.
   TRUFFLEHOG_RC=0
-  trufflehog filesystem . --only-verified --json \
+  timeout 300 trufflehog filesystem . --only-verified --json \
     > /tmp/vuln-scan/trufflehog.json 2>/dev/null || TRUFFLEHOG_RC=$?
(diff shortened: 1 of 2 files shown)


skills/vuln-scanner/SKILL.md
```diff
@@ -162,9 +162,14 @@ fi
 
 # --- Secrets: TruffleHog (only-verified = actually authenticates) ---
 if command -v trufflehog >/dev/null 2>&1; then
+  # BOTH passes are bounded. Run these verbatim; if you invoke trufflehog (or any
```
Svector-anu, line 165: [ISSUE] Refresh and commit `eyebrowlock.json` for this governed skill change — the required `verify` check rejects the current integrity drift, so this head cannot pass CI or merge as-is.

Svector-anu (commented):
The integrity lockfile must be refreshed before this head can pass required CI.

- [ISSUE] skills/vuln-scanner/SKILL.md:165 — refresh and commit `eyebrowlock.json`; the required `verify` check rejects the current integrity drift, so this head cannot pass CI or merge as-is.

PR #1073, opened 2026-09-17
feat(vuln-scanner): harden disclosure routing (A5.0 SECURITY.md-first, A5b.0 PVR preflight)

aaronjmars:
## What

Two additive disclosure-routing fixes for `vuln-scanner`, ported from the aeon-vuln instance where they were battle-tested. Both stop a confirmed code flaw from being filed on the wrong channel.

### A5.0 - SECURITY.md-first intake (new sub-step before the finding-type matrix)

GitHub PVR being *technically enabled* does not make it the correct channel. Many vendors run PVR org-wide but triage through their own PSIRT intake, so a GitHub PVR lands in a queue they never read and silently violates their stated process. A5.0 fetches the repo's policy (including the org-level `{owner}/.github` fallback), resolves the designated intake channel in precedence (vendor PSIRT / bug-bounty portal, then security email, then explicit GitHub PVR), and only then falls back to the existing finding-type matrix.

Prevents the real `google/agents-cli` misroute: a code flaw was filed as GitHub PVR `GHSA-x742-5676-qjpj` while the repo policy said `g.co/vulnz`.

### A5b.0 - PVR preflight on HTTP status code (new preflight atop A5b)

The old status read (`--jq '.enabled' 2>/dev/null`) is wrong on any non-200: `gh api --jq` prints the JSON error body, not empty, so a 404 (private / missing / renamed) or a 403/429 rate-limit was silently read as "PVR disabled" and misrouted a code flaw to email. A5b.0 reads the status code once and sets an explicit state: `enabled` / `disabled` / `unavailable` / `access-unknown`. Compose + POST only when `enabled`.

## Scope

Additive only. The legacy scan flow, the A4.5 PoC-verification gate, Arm D (poc-smoke), and the Riva/shadow kernel are untouched. `git diff --stat`: 44 insertions, 1 deletion, both hunks in the A5 routing section.

## Not included (deferred, coupled)

- A6.5 coverage manifest + the A3 scanner-metrics split depend on the instance's A3.5 entrypoint-inventory restructure, which would replace canon's A3.5 fuzz-harness + A3.6 agentic sections. Not a clean additive port.
- The A1 alternating feed picker depends on the `bigco-repos` skill, which canon does not ship.


Svector-anu (commented):

